PR #35, opened 2012-02-16
Support for JRuby + Pageant + Windows

arturaz:
Currently support is broken, because neither dl nor UNIXSocket works on windows.

This adds windows support for jruby.

### DIFF
--- a/net-ssh.gemspec
+++ b/net-ssh.gemspec
@@ -17,7 +17,7 @@
   # This has two flavours with java one actually doing something and other
   # one just raising error. This is a workaround for no ability to specify
   # platform specific dependencies in gemspecs.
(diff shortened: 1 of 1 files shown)


mfazekas:
@arturaz is this still required on java/win side? Doesn't Jruby support ```fiddle``` now?
I'm asking because java pageant reimplements to whole agent logic, while mri ruby windows pageant just implements the IO to pageant. This means features like agent_socket_factory is not supported by java_pageant.

I'm thinking about removing java_pageant from 4.0 and later releases, and just using the https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/authentication/pageant.rb implementation on both jruby and mri ruby.

arturaz:
No idea. Not using ruby right now, so I'm not really qualified to answer.

Issue #38, opened 2012-04-03
Remote and Local Forward


arion:
Hi!

Net::SSH.configuration_for ignore LocalForward and RemoteForward from ~/.ssh/config and Net::SSH.start has no option for forward

mfazekas:
We have remote and local forward programatically. 
https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh.rb#L47-L59
```ruby
  # == X == "forward connections on a local port to a remote host"
  #
  #   Net::SSH.start("host", "user", :password => "password") do |ssh|
  #     ssh.forward.local(1234, "www.google.com", 80)
  #     ssh.loop { true }
  #   end
  #
  # == X == "forward connections on a remote port to the local host"
  #
  #   Net::SSH.start("host", "user", :password => "password") do |ssh|
  #     ssh.forward.remote(80, "www.google.com", 1234)
  #     ssh.loop { true }
  #   end
```

i'd be interested in concrete usecase, LocalForward and RemotForward is more for interactive use. Even if we support those options it should be non default as for non interactive cases doing the forward might not be desirable.

Issue #25, opened 2011-08-22
Net::SSH doesn't work if a proxy takes a %r


icco:
Apparently, proxy commands, according to SSHd, can have a %r for where to pass the user name. Net::SSH doesn't support this and fails.

```
ruby-1.9.2-p290 :004 > Net::SSH.start('example.net', 'ec2-user') {|ssh| p ssh }
ArgumentError: unknown key: r
from /home/natwelch/.rvm/gems/ruby-1.9.2-p290/gems/net-ssh-2.2.0/lib/net/ssh/proxy/command.rb:46:in `block in open'
from /home/natwelch/.rvm/gems/ruby-1.9.2-p290/gems/net-ssh-2.2.0/lib/net/ssh/proxy/command.rb:37:in `gsub'
from /home/natwelch/.rvm/gems/ruby-1.9.2-p290/gems/net-ssh-2.2.0/lib/net/ssh/proxy/command.rb:37:in `open'
from /home/natwelch/.rvm/gems/ruby-1.9.2-p290/gems/net-ssh-2.2.0/lib/net/ssh/transport/session.rb:66:in `block in initialize'
from /home/natwelch/.rvm/rubies/ruby-1.9.2-p290/lib/ruby/1.9.1/timeout.rb:44:in `timeout'
from /home/natwelch/.rvm/rubies/ruby-1.9.2-p290/lib/ruby/1.9.1/timeout.rb:89:in `timeout'
from /home/natwelch/.rvm/gems/ruby-1.9.2-p290/gems/net-ssh-2.2.0/lib/net/ssh/transport/session.rb:66:in `initialize'
from /home/natwelch/.rvm/gems/ruby-1.9.2-p290/gems/net-ssh-2.2.0/lib/net/ssh.rb:186:in `new'
from /home/natwelch/.rvm/gems/ruby-1.9.2-p290/gems/net-ssh-2.2.0/lib/net/ssh.rb:186:in `start'
from (irb):4
from /home/natwelch/.rvm/rubies/ruby-1.9.2-p290/bin/irb:16:in `<main>'
```

This happens because my work puts the following in my ssh_config: `ProxyCommand /usr/sbin/prodssh --logtostderr --minloglevel=3 --dest_host=%h --dest_port=%p --dest_user=%r`

PR #139, opened 2014-01-14
Support %r in ProxyCommand configuration in ssh_config files as defined in OpenSSH.

yugui:
This is a fix of #25 in github, and #38 in lighthouse.
http://net-ssh.lighthouseapp.com/projects/36253/tickets/38-netssh-doesnt-work-if-a-proxy-takes-a-r

This commit changes the API of Net::SSH::Proxy#open().  Although this commit also let subclasses of Net::SSH::Proxy follow the change, please let me know if you have any other concern.

### DIFF
--- a/lib/net/ssh/transport/session.rb
+++ b/lib/net/ssh/transport/session.rb
@@ -64,7 +64,13 @@ module Net; module SSH; module Transport
 
       debug { "establishing connection to #{@host}:#{@port}" }
       factory = options[:proxy] || TCPSocket
-      @socket = timeout(options[:timeout] || 0) { @bind_address.nil? || options[:proxy] ? factory.open(@host, @port) : factory.open(@host,@port,@bind_address) }
+      @socket = timeout(options[:timeout] || 0) {
+        case
+        when options[:proxy] then factory.open(@host, @port, options)
+        when @bind_address.nil? then factory.open(@host, @port)
+        else factory.open(@host, @port, @bind_address)
+        end
+      }
       @socket.extend(PacketStream)
       @socket.logger = @logger
 
--- a/lib/net/ssh/proxy/command.rb
+++ b/lib/net/ssh/proxy/command.rb
@@ -33,13 +33,20 @@ module Net; module SSH; module Proxy
 
     # Return a new socket connected to the given host and port via the
     # proxy that was requested when the socket factory was instantiated.
-    def open(host, port)
+    def open(host, port, connection_options = nil)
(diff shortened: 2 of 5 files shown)


lib/net/ssh/transport/session.rb
```diff
@@ -64,7 +64,13 @@ def initialize(host, options={})
 
       debug { "establishing connection to #{@host}:#{@port}" }
       factory = options[:proxy] || TCPSocket
-      @socket = timeout(options[:timeout] || 0) { @bind_address.nil? || options[:proxy] ? factory.open(@host, @port) : factory.open(@host,@port,@bind_address) }
+      @socket = timeout(options[:timeout] || 0) {
+        case
+        when options[:proxy] then factory.open(@host, @port, options)
```
mfazekas: as noted by https://github.com/net-ssh/net-ssh/issues/220 ```options[:proxy]``` can't be both a factory class and dictionary. 
```factory = options[:proxy] || TCPSocket```
```factory.open(@host, @port, options)```

TiagoCardoso1983:
I think that, for this to make sense, all proxy should abide to the original TCPSocket.open method signature, otherwise this will break other proxy implementations that rely on this very signature. As per documentation:

> With no associated block, IO.open is a synonym for ::new. If the optional code block is given, it will be passed io as an argument, and the IO object will automatically be closed when the block terminates. In this instance, ::open returns the value of the block.
http://www.ruby-doc.org/core-1.9.3/IO.html#method-c-open

open should therefore also be implemented as a class method, as the initializer will anyway open the socket. Besides that, given that the least common denominator here will be the class TCPSocket, the API should at least mimick it (I would even go as far as suggesting that all of them should "be" a TCPSocket, that is, inherit from it).

mfazekas:
* Pre 2.8 implementations supported TCPSocket and TCPSocket like classes as proxy. As classes passed to ```:proxy```  needed signature with ```open(host,port)```
* 2.8 and 2.9 requires the classes passed to ```:proxy``` to have open with signature ```open(host,port,options={})```

I think the code is confusing as it says:
```factory = options[:proxy] || TCPSocket```
so you could assume ```:proxy``` can be a TCPSocket. I don't think this is documented anywhere. The documentation says that ```Proxy::``` classes are supported.
see https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh.rb#L161 ```:proxy => a proxy instance (see Proxy) to use when connecting``` 

I don't agree that proxy should be a TCPSocket subclass.

Issue #124, opened 2013-09-26
ssh-rsa-cert-v01@openssh.com ssh key type support


vkhatri:
Hello,

I am using ssh signed public key + private key authentication but net-ssh does not have implementation for signed public key.

exception:
/usr/lib/ruby/gems/1.8/gems/net-ssh-2.6.7/lib/net/ssh/buffer.rb:269:in `read_keyblob': unsupported key type`ssh-rsa-cert-v01@openssh.com' (NotImplementedError)

I would not know the implementation of it, but will be glad to help in testing.

-V

PR #142, opened 2014-01-21
Handle ssh-rsa and ssh-dss certificate files

bobveznat:
This does not implement certificate based authentication (described here
http://www.openbsd.org/cgi-bin/cvsweb/src/usr.bin/ssh/PROTOCOL.certkeys?rev=HEAD)
but instead makes it so that if the certificate is not needed for
authentication net-ssh doesn't cause the entire application to die.

The net-ssh test suite continues to pass. On my own machine I did tests
with certificates loaded and verified that although authentication could
not proceed to a host requiring a certificate it at least did not die.

I also verified that I can continue to use normal rsa and dsa keys to
ssh to hosts that do not require certificates even when the certificates
are loaded into my ssh-agent instance.

This is a potential solution to issue #124 and an alternative to the one
presented in pull request #134.

### DIFF
--- a/lib/net/ssh/buffer.rb
+++ b/lib/net/ssh/buffer.rb
@@ -243,14 +243,14 @@ module Net; module SSH
     # a key. Only RSA, DSA, and ECDSA keys are supported.
     def read_keyblob(type)
       case type
-        when "ssh-dss"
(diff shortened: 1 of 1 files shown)


v6:
// , Someone needs to update the notes for https://pypi.python.org/pypi/ssh-ca/0.3.2.

Issue #168, opened 2014-06-18
2.9.1 - Proxy command on Windows 7: 'read_nonblock': Bad file descriptor (Errno::EBADF)


bvanheu:
Hi,

I'm using Net-ssh 2.9.1 with Ruby 2.0.0p353 32-bit on Windows 7 32-bit.

I'm trying to use a proxy command but net-ssh fails with "'read_nonblock': Bad file descriptor (Errno::EBADF)".

Here is the code i'm using:

```
require 'rubygems'
require 'net/ssh'
require 'net/ssh/proxy/command'

connect_opts = {
  auth_methods:          ["none", "publickey", "hostbased", "password"],
  config:                false,
  forward_agent:         false,
  keys:                  ["C:/Users/user/.vagrant.d/insecure_private_key"],
  keys_only:             true,
  paranoid:              false,
  password:              nil,
  port:                  22,
  timeout:               15,
  user_known_hosts_file: [],
  verbose:               :debug,
  proxy:                 Net::SSH::Proxy::Command.new("ssh '1.2.3.4' -l 'user' nc %h %p")
}

Net::SSH.start('192.168.121.171', 'vagrant', connect_opts) do|ssh|
  ssh.exec('ls') do|ch, stream, data|
    puts "Got this on the #{stream} stream: "
    puts data
  end
end
```

And the corresponding stack trace:

```
$ ./ruby.exe c:/Users/user/Desktop/testssh.rb
D, [2014-06-18T18:22:19.465400 #1324] DEBUG --net.ssh.transport.session[3da988]: establishing connection to 192.168.121.171:22
D, [2014-06-18T18:22:19.902200 #1324] DEBUG -- net.ssh.transport.session[3da988]: connection established
I, [2014-06-18T18:22:19.902200 #1324]  INFO -- net.ssh.transport.server_version[3d7b20]: negotiating protocol version
D, [2014-06-18T18:22:19.902200 #1324] DEBUG -- net.ssh.transport.server_version[3d7b20]: remote is `SSH-2.0-OpenSSH_6.6p1 Ubuntu-2ubuntu1'
D, [2014-06-18T18:22:19.902200 #1324] DEBUG -- net.ssh.transport.server_version[3d7b20]: local is `SSH-2.0-Ruby/Net::SSH_2.9.1 i386-mingw32'
[...]

PR #169, opened 2014-06-27
Use sysread and syswrite on Windows instead of read_nonblock and write_n...

marc-etienne:
Use sysread and syswrite on Windows instead of read_nonblock and write_nonblock in proxy command pipe (#168)


mfazekas:
Thanks for the patch sorry for taking it so long to merge

PR #216, opened 2015-01-10
Stop appending algorithms the user hasn't configured

voidus:
By appending these algorithms, the user configuration is used as a preference
list instead of a white list.
This means that a connection might use a weaker algorithm than expected.

### DIFF
--- a/lib/net/ssh/transport/algorithms.rb
+++ b/lib/net/ssh/transport/algorithms.rb
@@ -223,11 +223,6 @@ module Net; module SSH; module Transport
               supported
             end
             lwarn { "unsupported #{algorithm} algorithm: `#{unsupported}'" } unless unsupported.empty?
-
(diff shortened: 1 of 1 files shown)


mfazekas:
Thanks for the PR. I agree with it but we need to be careful if the reading of config is not correct it's possible that this change will break some users.
Tests needs to be fixed.

voidus:
I'll look into the tests.

Well, yeah. This might break some user's setup, but those setups are misconfigured. I just want to point out that we are adding `none` to the list of accepted ciphers, which is probably unacceptable for most situations.

mfazekas:
I meant that this "feature" could have hidden bugs in config parsing, which can surface now.

mfazekas:
#231 fixes the tests, and provides an option to append algorithms

PR #217, opened 2015-01-10
Check if SSH key is a file

kiela:
For one of servers which I have access to I need to use custom SSH key. I'm not using my default SSH key, so I made a custom entry in my `~/.ssh/config` to handle that. I specified value for IdentityFile pointing to correct id_rsa file. I am able to ssh to the server with no issues at all. However, when I try to deploy using Capistano, it fails with an error `connection failed for: dev.domain.com (Errno::EISDIR: Is a directory @ io_fread - /home/kiela/workspace/ruby/project`. I started digging in Capistrano but found nothing which can cause this. I moved to Net::SSH and noticed that when no key was specified in options for `Net::SSH.start` it somehow sets keys to current directory and tries to open that directory as a file. `File.readable?` returns `true` as directory can be read but the `KeyFactor.load_public_key` tries to read it and raises an `Errno::EISDIR` exception.
After checking if a key is a file and can be read, everything starts working like a charm.

I assume that there could be different issue with this, but we should check if specified SSH key is a file before we try to read it content.

### DIFF
--- a/lib/net/ssh/authentication/key_manager.rb
+++ b/lib/net/ssh/authentication/key_manager.rb
@@ -188,9 +188,9 @@ module Net
         def prepare_identities_from_files
           key_files.map do |file|
             public_key_file = file + ".pub"
-            if File.readable?(public_key_file)
-              { :load_from => :pubkey_file, :file => file }
-            elsif File.readable?(file)
+            if File.file?(public_key_file) and File.readable?(public_key_file)
+              { :load_from => :pubkey_file, :file => public_key_file }
+            elsif File.file?(file) and File.readable?(file)
               { :load_from => :privkey_file, :file => file }
             end
           end.compact
@@ -209,7 +209,7 @@ module Net
             begin
               case identity[:load_from]
               when :pubkey_file
-                key = KeyFactory.load_public_key(identity[:file] + ".pub")
+                key = KeyFactory.load_public_key(identity[:file])
                 { :public_key => key, :from => :file, :file => identity[:file] }
               when :privkey_file
                 private_key = KeyFactory.load_private_key(identity[:file], options[:passphrase], ask_passphrase)
--- a/test/authentication/test_key_manager.rb
+++ b/test/authentication/test_key_manager.rb
@@ -156,7 +156,9 @@ module Authentication
 
       def stub_file_private_key(name, key, options = {})
         manager.add(name)
+        File.stubs(:file?).with(name).returns(true)
         File.stubs(:readable?).with(name).returns(true)
+        File.stubs(:file?).with(name + ".pub").returns(true)
         File.stubs(:readable?).with(name + ".pub").returns(false)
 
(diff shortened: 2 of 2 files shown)


lib/net/ssh/authentication/key_manager.rb
```diff
@@ -209,8 +209,8 @@ def load_identities(identities, ask_passphrase)
             begin
               case identity[:load_from]
               when :pubkey_file
-                key = KeyFactory.load_public_key(identity[:file] + ".pub")
-                { :public_key => key, :from => :file, :file => identity[:file] }
+                key = KeyFactory.load_public_key(identity[:file])
+                { :public_key => key, :from => :file, :file => identity[:file]}
```
mfazekas: I don't think this space removal was intentional

mfazekas:
Thanks for the PR! Can you explain what bug fixed under what conditions?

kiela:
@mfazekas I'm sorry, I forgot about describing the issue. I have updated the description.

lib/net/ssh/authentication/key_manager.rb
kiela: No it wasn't. Fixed.

mfazekas:
Thanks!

hairihan:
Hi @kiela ,

I have meet a same problem as you described, but I'm not understand well about your solution.
I have set `my_key.pem` to readable, but it still can not connect the server.

Here is ssh files content in my workstation. and may I ask for a advice or tips?

screen shot 0027-10-12 at 16 27 59

kiela:
hey @hairihan 

My issue was that I was using custom SSH key for connecting to a server. I specified that key in my SSH config file so I was able to ssh to server's console. However, I didn't specified anything in Net::SSH configuration so Net::SSH was using a value kept in `public_key_file` variable as a SSH key - which in my case was my current directory AFAIR. That behaviour caused Net::SSH to crash while trying to establish SSH connection. My solution was to make Net::SSH to use value from `public_key_file` variable if it's a file, if not - use value from `file` variable and if neither of them exist - do nothing which makes SSH to use either value specified in `~/.ssh/config` or default value.

What I can advise you is to take a step back and start checking one part at a time at which point SSH connection is not established:
 - make sure you can just ssh to your server by specifying SSH key in command line
 - make sure you can just ssh to your server using configuration in your `~/.ssh/config` file without specifying SSH key in command line
 - if you can ssh to your server from command line, try to configure Net::SSH to use your SSH key
 - if Net::SSH works as expected after specifying SSH key in configuration - there still might be a bug in the place which I made a PR to
 - if Net::SSH still doesn't work, I assume that there might be another use case which you have just came across so more information will be needed.

hairihan:
Hi @kiela 
Thank you and sorry for late response.
I have find that it was caused by rubber finally, just a little different from you.
But your advise helped me a lots~

Thanks you again~

kiela:
@hairihan Cool! I'm glad I could help!

PR #221, opened 2015-01-24
Switch pageant.rb to use Fiddle

cfis:
pageant.rb no longer works on Ruby 2.2 on Windows since the dl module has been removed.  This patch updates pageant.rb to use fiddle instead for ruby 2.2 and higher.  Pageant.rb tests all pass.

### DIFF
--- a/lib/net/ssh/authentication/pageant.rb
+++ b/lib/net/ssh/authentication/pageant.rb
@@ -9,6 +9,12 @@ else
   require 'fiddle'
   require 'fiddle/types'
   require 'fiddle/import'
+
+  # For now map DL to Fiddler versus updating all the code below
+  module DL
+    CPtr = Fiddle::Pointer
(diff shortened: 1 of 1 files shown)


mfazekas:
looks good, thanks!

Have you also tested with an app too (capistrano, chef, puppet)?

cfis:
Not yet, but in progress with Capistrano.  Needed to first get this issue fixed since it blew up cap install (upgrading from capistrano 2.x to 3.x).

mfazekas:
I've started a doc describing how you can test your modified net-ssh with capistrano:
https://github.com/mfazekas/net-ssh/blob/testing_docs/CONTRIBUTING.md

Can you please check out and try it to run capistrano with the modified version? Thanks!

cfis:
I can confirm this now works using Capistrano on Windows 8.1.

PR #228, opened 2015-02-15
Ed25519 support

mfazekas:
No support for encrypted (aka password protected) keys yet. They works from non encrypted or if they are loaded from agent.

For encrypted keys we'd need a native library bcrypt_hash and bcrypt_pbkdf impl accessible from ruby.


blackwinter:
May I ask, what's the status of this? With this patch applied on top of 3.0.2, net-ssh successfully uses my ed25519 keys.

mfazekas:
@blackwinter i think i was stuck adding support for reading password protected key files, they can be only loaded from agent but not from file directly, should work if your keyfile is not password protected

blackwinter:
Indeed, mine isn't ;) Any chance you might get "unstuck" soon? Anything I can do to help? It's not utterly pressing for me, seeing as this patch applies cleanly and does the job. But out-of-the-box support for ed25519 would be nice - even without the ability to read password-protected key files.

PR #230, opened 2015-02-22
Add Rubinius to Build Matrix with Allowed Failure

bjfish:
Please add Rubinius to the build matrix with failure allowed.

I am trying to add this to ensure this Gem works in Rubinius and to complete this dashboard: http://bjfish.github.io/rubinius-gem-build-status/page2/.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,4 +17,3 @@ matrix:
   allow_failures:
     - rvm: jruby-head
     - rvm: 1.8.7
-    - rvm: rbx-2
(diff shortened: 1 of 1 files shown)


mfazekas:
Thanks! I think we can just add rbx-2 without allowed_failure for now as it's passing. Once it fails in a way it's not trivial to fix, we'll move to allowed_failure.

bjfish:
@mfazekas I've updated the PR to remove the allowed failure, thanks!

mfazekas:
Thanks!

PR #238, opened 2015-04-21
Fix ArgumentError: Could not parse PKey: no start line

Antiarchitect:
Capistrano 2.15.5 and net-ssh 2.9.3.beta1 and standart linux pub/pkey locations:
~/.ssh/id_rsa
~/.ssh/id_rsa.pub

Capistrano fails with ArgumentError: Could not parse PKey

After some time of exploration I've found that for this situation public key is checked first and :load_from => :pubkey_file identity appears. After that I'm totally cannot understand how :pubkey_branch works in load_identities - there is no setting of private key at all. So it's just a workaround for my situation, but it definetly can be done better with help of people that know the internals.

### DIFF
--- a/lib/net/ssh/authentication/key_manager.rb
+++ b/lib/net/ssh/authentication/key_manager.rb
@@ -187,11 +187,10 @@ module Net
         # Prepares identities from user key_files for loading, preserving their order and sources.
         def prepare_identities_from_files
           key_files.map do |file|
-            public_key_file = file + ".pub"
-            if readable_file?(public_key_file)
-              { :load_from => :pubkey_file, :file => public_key_file }
-            elsif readable_file?(file)
+            if readable_file?(file)
               { :load_from => :privkey_file, :file => file }
+            elsif readable_file?(public_key_file = file + ".pub")
+              { :load_from => :pubkey_file, :file => public_key_file }
             end
           end.compact
         end


mfazekas:
Thanks, for PR, let me dig into what's going on.

mfazekas:
1.) i've managed to reproduce the problem at integration test see: https://github.com/net-ssh/net-ssh/blob/master/test/integration/test_id_rsa_keys.rb#L20 
2.) reason we deal with pubkey if they are available, is that while we can get public key from the private key too, it might ask a password. We need pubkeys for filering out keys already in agent.

So your change breaks in this situation: your key is in the agent, and you have it available on standard location too. 
* Before your change net-ssh did found the pubkey recognized that this key is in agent and used the agent version if needed. 
* After your changes if the privkey is encrypted then we'll ask for password for it we'll load the privkey then, we'll recognize that the same key is in the agent.

mfazekas:
@Antiarchitect it would be great if you could verify my patch for the issue

mfazekas:
@Antiarchitect i've released https://rubygems.org/gems/net-ssh/versions/2.10.0.beta1 which should fix the issue please verify when you have a chanche

xn:
It does not fix the issue for me. Is there a version I can rollback to?

mfazekas:
@xn what is you issue? Can you please post more details - like stacktrace, and your setup?

```gem uninstall net-ssh -v 2.10.0.beta1``` should remove 2.10.0.beta1

xn:
Sorry for the false alarm, the problem was with capistrano. For whatever reason, my `ssh_options[:keys]` configuration was no longer being parsed correctly. Removing the `keys` key and value from the hash, fixed it for me.

Issue #235, opened 2015-04-02
Ruby 2.2 64bit / Pageant / Windows / "bignum too big to convert into `long'"


ccoenen:
This issue is exclusive to the **64 bit** build of ruby 2.2.1. The 32bit build runs fine.

The current master (https://github.com/net-ssh/net-ssh/commit/3ed5b1963c1aafbb1ec239b5a0a20cc57ea92194) won't work on ruby 2.2.1-64 / Windows. It fails with this message:

```
SSHKit::Runner::ExecuteError: Exception while executing on host --redacted--: bignum too big to convert into `long'
C:/Tools/Ruby22-x64/lib/ruby/gems/2.2.0/bundler/gems/net-ssh-3ed5b1963c1a/lib/net/ssh/authentication/pageant.rb:191:in `malloc'
C:/Tools/Ruby22-x64/lib/ruby/gems/2.2.0/bundler/gems/net-ssh-3ed5b1963c1a/lib/net/ssh/authentication/pageant.rb:191:in `malloc_ptr'
C:/Tools/Ruby22-x64/lib/ruby/gems/2.2.0/bundler/gems/net-ssh-3ed5b1963c1a/lib/net/ssh/authentication/pageant.rb:255:in `get_token_information'
C:/Tools/Ruby22-x64/lib/ruby/gems/2.2.0/bundler/gems/net-ssh-3ed5b1963c1a/lib/net/ssh/authentication/pageant.rb:228:in `get_current_user'
C:/Tools/Ruby22-x64/lib/ruby/gems/2.2.0/bundler/gems/net-ssh-3ed5b1963c1a/lib/net/ssh/authentication/pageant.rb:204:in `get_security_attributes_for_user'
C:/Tools/Ruby22-x64/lib/ruby/gems/2.2.0/bundler/gems/net-ssh-3ed5b1963c1a/lib/net/ssh/authentication/pageant.rb:347:in `send_query'
C:/Tools/Ruby22-x64/lib/ruby/gems/2.2.0/bundler/gems/net-ssh-3ed5b1963c1a/lib/net/ssh/authentication/pageant.rb:324:in `send'
C:/Tools/Ruby22-x64/lib/ruby/gems/2.2.0/bundler/gems/net-ssh-3ed5b1963c1a/lib/net/ssh/authentication/agent/socket.rb:148:in `send_packet'
```

There has been some work towards 2.2 compatibility in January, so i wonder if this should actually be working or if it was not tested on 64 bit.

[...]

PR #239, opened 2015-04-23
Fixing pageant interaction on 64-bit platform 

pleaseking:
fix net-ssh not working with Pageant in 64-bit build. Problem was manually packing arrays into Win32 structures without taking alignment and 32bit/64bit field size discrepancies into account. 

Solved by using Fiddle Struct API - it packs correctly on both 32-bit and 64-bit system. 

Tested in both 32-bit and 64-bit Ruby (2.2.x, 2.1.x)

### DIFF
--- a/lib/net/ssh/authentication/pageant.rb
+++ b/lib/net/ssh/authentication/pageant.rb
@@ -153,6 +153,9 @@ module Net; module SSH; module Authentication
                                     'LPVOID Group', 'LPVOID Sacl',
                                     'LPVOID Dacl']
 
+      # The COPYDATASTRUCT is used to send WM_COPYDATA messages
+      COPYDATASTRUCT = struct ['uintptr_t dwData', 'DWORD cbData', 'LPVOID lpData']
+
       # Compatibility for security attribute retrieval.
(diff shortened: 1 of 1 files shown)


Antiarchitect:
Tested on Win x64 with Ruby 2.2.2 x64

pleaseking:
Also tested on Win x64 (Windows Server 2012 R2) with Ruby 2.1.6 x86.

mfazekas:
thanks, much :)

thargor:
Does stilll not work for me with, but now with the following error message on win 8.1 x64 ruby 2.1.5 and net:ssh v2.10.0.beta1.

    SSHKit::Runner::ExecuteError: Exception while executing as root@github.com: to_ptr should return a CPtr object
    DL::DLError: to_ptr should return a CPtr object

Should I open a new bug?

pleaseking:
I think yes, but can you check ruby 2.2.x ?

thargor:
It does connect using the agent with 2.2.1, but then runs into the "bignum too big to convert into `long'"  error from #235

PR #240, opened 2015-05-13
Fix NoMethodError `recv` in multithreaded usage of Net::SSH::Proxy::Command

mattbrictson:
This fixes a longstanding bug discovered by users of SSHKit: https://github.com/capistrano/sshkit/issues/150.

To summarize: when SSH connections are made in multiple threads using `ProxyCommand`, there is a small possibility that `Net::SSH.start` will fail with an error like this:

```
NoMethodError: undefined method `recv' for #<IO:fd 47>
```

I traced the problem down to Ruby itself. Apparently, defining singleton methods using this syntax (as is done by `proxy/command.rb`) is not thread safe:

```ruby
class << io
  def send(data, flag)
    write_nonblock(data)
  end

  def recv(size)
    read_nonblock(size)
  end
end
```

When executed in parallel by many threads, there is a small chance that the `recv` method will not be defined. As strange as this bug may seem, you can reproduce it consistently with this test case:

```ruby
require "thread"

def define_singleton_methods_in_thread
  Thread.new do
    100_000.times do
      obj = Object.new
      class << obj
        def one
        end
        def two
        end
        def three
        end
      end
      puts "FAIL" unless obj.respond_to?(:three)
    end
  end
end

# Test with 2 threads
2.times.map { define_singleton_methods_in_thread }.map(&:join)
```

On my machine this outputs `FAIL` three times.

The good news is that there is another way to define singleton methods in Ruby that *is* thread safe:

```ruby
# Not thread safe :(
class << obj
  def one
  end
  def two
  end
  def three
  end
end

# Thread safe :)
def obj.one
end
def obj.two
end
def obj.three
end
```

This PR rewrites changes the way `send` and `recv` are declared to use the latter syntax.

I plan to also open a bug report for MRI, but for now this should work around the problem.

### DIFF
--- a/lib/net/ssh/proxy/command.rb
+++ b/lib/net/ssh/proxy/command.rb
@@ -67,25 +67,23 @@ module Net; module SSH; module Proxy
         raise ConnectError, "#{e}: #{command_line}"
       end
       @command_line = command_line
-      class << io
-        if Gem.win_platform?
-          # read_nonblock and write_nonblock are not available on Windows
-          # pipe. Use sysread and syswrite as a replacement works.
-          def send(data, flag)
-            syswrite(data)
-          end
+      if Gem.win_platform?
+        # read_nonblock and write_nonblock are not available on Windows
+        # pipe. Use sysread and syswrite as a replacement works.
+        def io.send(data, flag)
+          syswrite(data)
+        end
 
-          def recv(size)
-            sysread(size)
-          end
-        else
-          def send(data, flag)
-            write_nonblock(data)
-          end
+        def io.recv(size)
+          sysread(size)
+        end
+      else
+        def io.send(data, flag)
+          write_nonblock(data)
+        end
 
-          def recv(size)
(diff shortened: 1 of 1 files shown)


mfazekas:
Thanks for the report, strange bug indeed.

myers:
@mfazekas any idea when this might make it into a released gem?  

Thanks for work on this.  It's something my coworkers and I run into on a daily basis.

mfazekas:
@myers there is now a beta release at: https://rubygems.org/gems/net-ssh/versions/2.10.0.beta1 containing the patch

mfazekas:
https://bugs.ruby-lang.org/issues/11153

willbryant:
Thanks for fixing this!

I've tested 2.10.beta2 with 100 rounds of connecting to 25 hosts simultaneously and had no errors.  For comparison, on average I hit this error 5 times doing that same test on 2.9.2.  Looking forward to the 2.10 release :).

PR #247, opened 2015-05-26
Use Socket.tcp with connect_timeout instead of TCPSocket+Timeout::timeout

carlhoerberg:


lib/net/ssh/transport/session.rb
```diff
@@ -82,7 +81,7 @@ def initialize(host, options={})
       @host_key_verifier = select_host_key_verifier(options[:paranoid])
 
 
-      @server_version = timeout(options[:timeout] || 0) { ServerVersion.new(socket, logger) }
+      @server_version = ServerVersion.new(socket, logger)
```
mfazekas: what's the reason for removing timeout from ServerVersion?

mfazekas:
Thanks for the PR looks good!

Can you give some background description, on why this change was neccesary? I assume the timeout will not abort a native TCP connect?

The connection_timeout is new in ruby 2.0. https://github.com/ruby/ruby/commit/5515afbf8f5b7e9f3c611691a9dd8326f7bd6f9c
I guess it's ok to be ruby 2.0 only as ruby 1.9 is EOL now.

carlhoerberg:
Exactly, `Timeout::timeout` starts a new thread, and let it sleep for the amount of seconds you set it to. So if you open 500 connections (like we are), you will end up with 1500 threads for the first "timeout" seconds. Using lower level semantics is much more resource efficient and safer (connections are properly cleaned up on timeout)  

http://blog.headius.com/2008/02/ruby-threadraise-threadkill-timeoutrb.html

lib/net/ssh/transport/session.rb
carlhoerberg: fixed, using `IO.select` instead of `Timeout`

PR #249, opened 2015-06-10
don't prompt for passphrases in non-interactive mode

jstanley0:
allow options[:passphrase] to be used with options[:non_interactive]


mfazekas:
:+1: thanks, looks good. I'd rename delay_passphrase to ignore_decryption_errors

PR #250, opened 2015-06-16
Add support for hostname hashes when searching for keys in known hosts

jefmathiot:
With the configuration option `HashKnownHost` enabled, OpenSSH now uses an HMAC-SHA1 hash to store hostnames in `known_hosts` files (this option is the default on Ubuntu).

Setting the `paranoid` option to `:secure` makes `net-ssh` fail to find the keys in the `known_hosts` file generated by OpenSSH as the `KnownHost` module only seeks for keys with hostnames stored as plain text.

As a side note, when storing hostnames as plain text, a single line may contain multiple hostnames, such as `github.com,100.100.100.100`. On the contrary, when using hostnames hashes OpenSSH creates a line  with a specific salt for each of the hostnames or IP addresses.

I added a second  `known_hosts` file for testing purposes with `|1|eKp+6E0rZ3lONgsIziurXEnaIik=|rcQB/rlJMUquUyFta64KugPjX4o=` as the hash. The `|1|` is called the "hash magic" and marks the value as hashed. The second part, up to the middle `|` character, is the salt. The last part is the HMAC-SHA1 of the entry with the salt used as the key. All values are Base64 encoded.

(Thanks for this useful gem, by the way)


lib/net/ssh/known_hosts.rb
```diff
@@ -127,6 +131,21 @@ def keys_for(host)
       keys
     end
 
+    # Indicates whether one of the entries matches an hostname that has been
+    # stored as a HMAC-SHA1 hash in the known hosts.
+    def known_host_hash?(hostlist, entries, scanner)
+      if hostlist.size == 1 && hostlist.first =~ /\A\|1(\|.+){2}\z/
```
carlhoerberg: why does the hostlist only have to have one entry?

jefmathiot: Because when the hostname is hashed, `hostlist = host.split(/,/) = ["|1|eKp+6E0rZ3lONgsIziurXEnaIik=|rcQB/rlJMUquUyFta64KugPjX4o="]`

carlhoerberg: I see, thanks.

mfazekas:
:+1: thanks for the patch

PR #265, opened 2015-08-18
Document that Net::SSH::Connection::Session#exec! returns nil if there is no output from the remote command

donv:
The docs state that #exec will return a string, and an empty string would be consistent.  Currently `nil` is returned, and this should be documented.

### DIFF
--- a/lib/net/ssh/connection/session.rb
+++ b/lib/net/ssh/connection/session.rb
@@ -351,8 +351,8 @@ module Net; module SSH; module Connection
     end
 
     # Same as #exec, except this will block until the command finishes. Also,
-    # if a block is not given, this will return all output (stdout and stderr)
-    # as a single string.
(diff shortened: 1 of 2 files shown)


donv:
@mfazekas Is it a conscious choice to return nil on empty command output?  Maybe we should change it to return an empty string?

carlhoerberg:
That would break compabilty for a lot of ppl. I for one know that we rely
on this (not so intuitive) behavior in many cases.

mfazekas:
It was not a conscious choice IMO, but an impl detail. I think it's ok to change this is an a major release if it's properly documented. @carlhoerberg what is your usecase do you compare result directly to nil?

mfazekas:
Actually will be revoking just released 3.0.0 gem and changing exec to return an empty string in 3.0

donv:
@mfazekas Thanks!

PR #267, opened 2015-08-22
re-raise Errno::ETIMEDOU from Socket.tcp as Net::SSH::ConnectionTimeout

carlhoerberg:
Keep backward compatibility with 2.9 at connection timeout exceptions.


carlhoerberg:
Oh, actually, `Timeout::Error` was raised before, but I think `Net::SSH::ConnectionTimeout` is much better then both `Errno::ETIMEDOU` and the too generic `Timeout::Error` . If server version negotation timed out `Net::SSH::ConnectionTimeout` was raised, even in 2.9, so ppl are probably (hopefully) already catching that exception.

PR #280, opened 2015-11-04
Send CHANNEL_CLOSE after output buffer is empty.

causton81:
My app creates three remote port forwards. I was seeing pretty regular ssh disconnects with errors similar to "Received data for nonexistent channel".  I tracked this down to the port forward logic. What happens is  this call to channel.close:

  https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/service/forward.rb#L304

will immediately send a CHANNEL_CLOSE even though there is still pending data in the output buffer.
When the window opens up and https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/connection/channel.rb#L312 comes around then data is sent after the CHANNEL_CLOSE.

There may be a cleaner way to fix this, but this patch seems to be working for me (the joy of race conditions). I try to keep processing the channel while its buffer has data and only send CHANNEL_CLOSE when the buffer is empty.

CHANNEL_EOF is already protected by an `if` that ensures the buffer is empty before sending.


lib/net/ssh/connection/channel.rb
```diff
@@ -497,6 +496,12 @@ def enqueue_pending_output #:nodoc:
             break
           end
         end
+
+        if @closing and not @sent_close and output.length < 1
```
mfazekas: What's the reason for putting this code here, vs process? In process we already seems to have similiar handling for EOF.

https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/connection/channel.rb#L310

mfazekas:
I'm not sure we want to wait for all data to be sent before sending close. Another approach is cleaning the output buffer after close. Do you know who calls close in your usecase?

lib/net/ssh/connection/channel.rb
causton81: Ok, it seems like the CLOSE could go here instead.

causton81:
The problem is here: https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/service/forward.rb#L304
In my case the remote forward is coming back to a squid proxy which is closing the connection after sending a response. From ssh perspective the channel is still open on both sides, net-ssh is closing it because the forwarded socket is closed.  I think discarding the output buffer in this case will discard a valid part of the HTTP response.

mfazekas:
So you're saying that the http server closes the connection right after sending response. Then https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/service/forward.rb#L304 causes the channel to close but we still want to transfer the possibly unsent fragments of the HTTP packet before closing the channel. Yes this makes sense.

It seems that on other places we call channel.close is also a gracefull close so yes your fix seems to be correct.

causton81:
Excellent. FWIW I have been using the patched version off and on over the last few weeks and it seems to be working well; I haven't noticed any issues.

mfazekas:
So we need to fix issues with tests see the travis failures, ideally we'd also need a testcase reproducing the problem

causton81:
Ok, I will try to square up the tests this week.

EYurchenko:
Looks like a fix for exactly what we are facing here https://github.com/chef/chef-provisioning/issues/446 but I just tested this branch and I am still seeing the error. Any thoughts are more than welcome. Thanks.

causton81:
@EYurchenko, chef-* gems are currently pinned to net-ssh ~> 2.x IIRC; it may be using an older stock net-ssh gem? I had to cherry-pick my changes over to the 2-10-release branch and I cut a 2.10.1.pre.1 tag locally. Try adding this to your :ssh_options:
```ruby
machine_options :ssh_options => {:logger => Chef::Log,}
```

then you should see net-ssh logs in your chef output (chef-client -l info ...)

pburkholder:
@causton81 If you have a PR to merge this fix to 2.10 then that might expedite getting this into Chef.

EYurchenko:
@causton81 you were right I was using master branch from your clone. With correct branch everything works as expected! We need that fix badly.

causton81:
@EYurchenko, excellent!

@pburkholder, I'll open a PR on 2-10-release which looks like the right branch.

lib/net/ssh/connection/channel.rb
```diff
@@ -274,7 +274,7 @@ def wait
     # invoked, but the server has not yet responded with a CHANNEL_CLOSE
     # packet of its own.
     def closing?
-      @closing
+      @closing and @sent_close
```
mfazekas: is this neccesary? only place we seems to be using this is https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/connection/session.rb#L223

maybe we should change that ```if closing?``` to ```closed?``` and implement ```closed``` as you've described.

lib/net/ssh/connection/channel.rb
```diff
@@ -311,10 +316,16 @@ def process
       @on_process.call(self) if @on_process
       enqueue_pending_output
 
-      if @eof and not @sent_eof and output.empty? and remote_id
+      if @eof and not @sent_eof and output.empty? and remote_id and not @sent_close
         connection.send_message(Buffer.from(:byte, CHANNEL_EOF, :long, remote_id))
         @sent_eof = true
       end
+
+      if @closing and not @sent_close and output.length < 1 and remote_id
```
mfazekas: please use ```output.empty?``` here as well

lib/net/ssh/connection/session.rb
```diff
@@ -583,9 +591,10 @@ def channel_close(packet)
         info { "channel_close: #{packet[:local_id]}" }
 
         channel = channels[packet[:local_id]]
+        channel[:remote_closed] = true
```
mfazekas: I'm a bit bothered that we're using ```:remote_closed``` as a property here vs closed which is an attribute. Maybe ```local_closed``` and ```remote_closed``` should be normal attributes and we should just deprecate old ```closing?``` What do you think?

mfazekas:
:clap: Thanks, i managed to reproduce the issue in an integration test at https://github.com/net-ssh/net-ssh/pull/289/files#diff-2b09e91e68b7981cddc37ff1d6fbe30bR341 and the changes did fix it.

lib/net/ssh/connection/session.rb
causton81: I pushed a new commit with remote_closed? and local_closed? methods.

causton81:
Thanks for doing the integration test and a bit of cleanup to get this merged.

tyler-ball:
Thanks for getting this fixed and merged!  Is there a planned release coming up that will include this?

mfazekas:
https://rubygems.org/gems/net-ssh/versions/3.0.2.rc1 and https://rubygems.org/gems/net-ssh/versions/2.9.4.rc1 should be fixing it.

sawanoboly:
@mfazekas Would it be possible for you to release 2.9.4 ?

mfazekas:
@sawanoboly have you tested 2.9.4.rc1, i'd prefer some feedback on it first

sawanoboly:
@mfazekas 

I've tested 2-9-release/HEAD with my tool. https://github.com/higanworks/knife-zero/issues/69

It resolves my issue and works fine.

mfazekas:
2.9.4 released

PR #282, opened 2015-11-11
Support using a default user(name)

### LINKED ISSUE #281: Net::SSH does not fall-back to using local user's username
We tried to connect to another host without setting the username, e.g:
```
require 'net/ssh'
Net::SSH.start(HOST, nil) do |ssh|
  ssh.exec!('env')
end
```
However, instead of it defaulting to trying to connect to said `HOST` with the local user's username, Net::SSH tried to connect with an empty username:
```
Text will be echoed in the clear. Please install the HighLine or Termios libraries to suppress echoed text.
@HOST's password:
@HOST's password:
@HOST's password:
Net::SSH::AuthenticationFailed: Authentication failed for user @HOST
	from /usr/local/rvm/gems/ruby-2.2.1@kiwiops/gems/net-ssh-3.0.1/lib/net/ssh.rb:239:in `start'
	from (irb):2
	from /usr/local/rvm/rubies/ruby-2.2.1/bin/irb:11:in `<main>'
2.2.1
```

We don't normally have an `~/.ssh/config` file, but did discover it will connect correctly with the following added:
```
Host *
  User USER
```

Is this expected behaviour?
If so:  we know we could use `$ENV['USER']` in the `Net::SSH.start()` (or similar system calls) is there any other way to get Net::SSH to default/fall-back to the local user's username?

Thanks

Will

NB: `HOST` and `USER` have been changed to protect the innocent.

mexisme:
FIXES #281

### DIFF
--- a/lib/net/ssh/authentication/session.rb
+++ b/lib/net/ssh/authentication/session.rb
@@ -51,6 +51,9 @@ module Net; module SSH; module Authentication
     # Attempts to authenticate the given user, in preparation for the next
     # service request. Returns true if an authentication method succeeds in
     # authenticating the user, and false otherwise.
+    #
+    # NB: If +username+ is +nil+ here, it will be implicitly converted to a
+    #     "" username, rather than the local (default) username.
     def authenticate(next_service, username, password=nil)
       debug { "beginning authentication of `#{username}'" }
 
--- a/lib/net/ssh.rb
+++ b/lib/net/ssh.rb
@@ -188,7 +188,7 @@ module Net
     #   password auth method
     # * :non_interactive => non interactive applications should set it to true
     #   to prefer failing a password/etc auth methods vs asking for password
-    def self.start(host, user, options={}, &block)
+    def self.start(host, user=nil, options={}, &block)
       invalid_options = options.keys - VALID_OPTIONS
       if invalid_options.any?
         raise ArgumentError, "invalid option(s): #{invalid_options.join(', ')}"
@@ -222,7 +222,7 @@ module Net
       transport = Transport::Session.new(host, options)
       auth = Authentication::Session.new(transport, options)
 
-      user = options.fetch(:user, user)
+      user = options.fetch(:user, user) || Etc.getlogin
       if auth.authenticate("ssh-connection", user, options[:password])
         connection = Connection::Session.new(transport, options)
         if block_given?
--- a/lib/net/ssh.rb
+++ b/lib/net/ssh.rb
@@ -98,6 +98,10 @@ module Net
     #     ssh.loop
     #   end
     #
+    # NB: If the +user+ parameter is omitted or nil, then the username of the
+    #     process used to invoke Net::SSH.start is used, instead (via
+    #     Etc.getlogin).
+    #
     # This method accepts the following options (all are optional):
     #
     # * :auth_methods => an array of authentication methods to try
--- a/test/start/test_user_nil.rb
+++ b/test/start/test_user_nil.rb
@@ -1,0 +1,27 @@
+require 'common'
+require 'net/ssh'
+
+module NetSSH
+  class TestStartUserNil < Test::Unit::TestCase
+    def setup
+      @authentication_session = mock('authentication_session')
+      Net::SSH::Authentication::Session.stubs(:new).returns(@authentication_session)
+      Net::SSH::Transport::Session.stubs(:new).returns(mock('transport_session'))
+      Net::SSH::Connection::Session.stubs(:new).returns(mock('connection_session'))
+    end
+
+    def test_start_should_accept_nil_user
+      @authentication_session.stubs(:authenticate).returns(true)
+      assert_nothing_raised do
+        Net::SSH.start('localhost')
+      end
+    end
+
+    def test_start_should_use_default_user_when_nil
+      @authentication_session.stubs(:authenticate).with() {|_next_service, user, _password| user == Etc.getlogin }.returns(true)
+      assert_nothing_raised do
+        Net::SSH.start('localhost')
+      end
+    end
+  end
+end


lib/net/ssh/authentication/session.rb
```diff
@@ -43,6 +46,7 @@ def initialize(transport, options={})
       @transport = transport
 
       @auth_methods = options[:auth_methods] || Net::SSH::Config.default_auth_methods
+      @default_user = options[:default_user] || Etc.getlogin
```
mexisme: I was unsure whether the preference is reading `Etc.getlogin` directly, or somehow via a similar `Net::SSH::Config` setting?

mfazekas:
I'm not sure about default_user case option. What is the usecase of 

```Net::SSH.start('localhost', nil,{default_username:'foo'}) ```
vs
```Net::SSH.start('localhost', 'foo')```

??

**Comment by mexisme** [comment]
> I'm not sure about default_user case option. What is the usecase of
> 
> Net::SSH.start('localhost', nil, {default_username:'foo'})
> vs
> Net::SSH.start('localhost', 'foo')
> 
> ??

It was a little tricky for me to follow how the default user _should_ be set:  when using the `ssh` command, it will decide the default user based on things like `~/.ssh/config` as well as the client-local username.
Should we be emulating this behaviour, or something slightly different, for example?
So, I erred on the side of flexibility for the first attempt, thinking it would be much easier to enable these other types of behvaiour if they were needed.

Also, some of the other Net::SSH options suggested that it was appropriate to allow additional overrides.

i.e. https://github.com/net-ssh/net-ssh/pull/282/files#diff-24b86dee68a1ab601927a977263f3eb2R49:
> I was unsure whether the preference is reading Etc.getlogin directly, or somehow via a similar Net::SSH::Config setting?


I overlooked explaining that, I'm sorry.   :-/

**Comment by mfazekas** [comment]
One argument is keeping this as simple as possible. The other is that this feature is not really usefull unless we support 
```
Host example
   User foo
Host example2
   User bar
```

After all we could just ```Etc.getlogin``` in Net::SSH.start. But looking at code we already seems to support that. We just need to default to Etc.getlogin when options(:user) or passed in user is nil.
I'm thinking on changing 
https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh.rb#L225
to
```ruby
user = options.fetch(:user, user) || Etc.getlogin
```

mexisme:
> One argument is keeping this as simple as possible.

Sounds good.

> After all we could just Etc.getlogin in Net::SSH.start. But looking at code we already seems to support that.

Thanks, that bit was slightly unclear (and how/where it was supported).

> We just need to default to Etc.getlogin when options(:user) and the passed in user is nil.
> I'm thinking on changing 
> https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh.rb#L225 to
> 
> `user = options.fetch(:user, user) || Etc.getlogin`

I wondered that, but since the below line supports default password style, it seemed that was the better place to support default username, as well.
(at https://github.com/net-ssh/net-ssh/pull/282/files#diff-24b86dee68a1ab601927a977263f3eb2L54):
```
def authenticate(next_service, username, password=nil)
```


i.e. change (at https://github.com/net-ssh/net-ssh/pull/282/files#diff-24b86dee68a1ab601927a977263f3eb2R60):
```
  username ||= default_user
```
to:
```
username ||= Etc.getlogin
```

mfazekas:
>I wondered that, but since the below line supports default password style, it seemed that was the better place to support default username, as well.
>(at https://github.com/net-ssh/net-ssh/pull/282/files#diff-24b86dee68a1ab601927a977263f3eb2L54):
>
>def authenticate(next_service, username, password=nil)

I don't think  your interpretation of  ```password=nil``` as default password style is correct. For me it more reads as  "password doesn't make sense for some of the authentication methods - like hostbased and publickey".

We'll also need to revise places where the user is defauled to ```ENV["USER"]``` like:
https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/authentication/methods/hostbased.rb#L36

mexisme:
> I don't think your interpretation of password=nil as default password style is correct. For me it more reads as "password doesn't make sense for some of the authentication methods - like hostbased and publickey".

Is there any liklihood that someone would want to use the Session class directly?
If so (and it looked like this might be true) then it doesn't feel right to make those users to deal with default behaviour.

> We'll also need to revise places where the user is defaulted to ENV["USER"] like:
> https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/authentication/methods/hostbased.rb#L36

```
            def authenticate_with(identity, next_service, username, key_manager)
              debug { "trying hostbased (#{identity.fingerprint})" }
              client_username = ENV['USER'] || username
```

That's an unexpected bit of code.  Why is that being done?
On most Unix systems, that would mean the passed-in `username` would always be ignored.

mfazekas:
> Is there any liklihood that someone would want to use the Session class directly?
> If so (and it looked like this might be true) then it doesn't feel right to make those users to deal with default behaviour.

I don't think many people would use Net::SSH::Authentication::Session directly.  I also don't think they would be much better with 
```auth.authenticate("ssh-connection", user_might_be_nil)``` vs  calling ```auth.authenticate("ssh-connection", user_might_be_nil || Etc.getlogin)```
I'd personally favour doing the defaulting at the caller and non changing Net::SSH::Authentication::Session, just because that looks a smaller change, and there is no significant advantage to the other version.

mexisme:
> I don't think many people would use Net::SSH::Authentication::Session directly. I also don't think they would be much better with  auth.authenticate("ssh-connection", user_might_be_nil) vs calling auth.authenticate("ssh-connection", user_might_be_nil || Etc.getlogin)

My only thought is that this doesn't reflect default SSH behaviour, so might be surprising to some.
I may add a small comment (or doc-comment?) mentioning this, if that's OK.

>I'd personally favour doing the defaulting at the caller and non changing Net::SSH::Authentication::Session, just because that looks a smaller change, and there is no significant advantage to the other version.

OK.  I've updated the PR to reflect this -- no tests, yet, but I'm writing them.

Thanks for talking this through with me.

mexisme:
> > I'd personally favour doing the defaulting at the caller and non changing Net::SSH::Authentication::Session, just because that looks a smaller change, and there is no significant advantage to the other version.
> 
> OK. I've updated the PR to reflect this -- no tests, yet, but I'm writing them.

Tests written.
Doc-comments updated.

NB:  I rebased the original PR to remove the previous code, to try to avoid polluting the canonical code with irrelevant commits.

mexisme:
Just wondering if you're happier with this PR, now, @mfazekas ?
Thanks.

mfazekas:
:+1:  Thanks much, i've merged it.

Issue #184, opened 2014-09-24
Net::SSH::Test not setup to handle request_pty and shell request


francisluong:
http://net-ssh.github.io/ssh/v2/api/classes/Net/SSH/Test/Channel.html

Suggest you add two convenience methods to Net::SSH::Test::Channel

#sends_pty_req - to correspond with Net::SSH::Channel#request_pty 
e.g. ```session.sends_channel_request(channel, "pty-req", true, nil, true)```

#sends_shell_req - to correspond with Net::SSH::Channel#send_channel_request("shell")
e.g. ```session.sends_channel_request(channel, "shell", true, nil, true)```

Also, lib/ssh/test/packet.rb will need to be updated to handle the new request packets - e.g.:
```
when CHANNEL_REQUEST
          parts = [:long, :string, :bool]
          case @data[1]
            when "exec", "subsystem" then parts << :string
            when "pty-req", "subsystem" then parts << :string
            when "shell", "subsystem" then parts << :string
```

mfazekas:
Both PTY and shell added, thanks for the report + PR. I've also added a hook to register any channel request not yet supported. https://github.com/net-ssh/net-ssh/blob/master/test/test/test_test.rb#L58-L74

PR #283, opened 2015-11-17
Add Net::SSH::Test::Channel#sends_request_pty for scripting PTY requests

p0deje:
Fixes net-ssh/net-ssh#184

It does not support scripting custom PTY request arguments, but I believe it's useful even in this state. Since I haven't found any tests for `Net::SSH::Test`, I haven't added any, but tested this manually.


mfazekas:
:+1: I would be more comfortable merging it if we'd had a testcase calling those.

Issue #269, opened 2015-09-03
socket being closed in forward.rb causing problem with gateway session 


thechile:
Hi

This is some info related to [https://github.com/net-ssh/net-ssh-gateway/issues/4](https://github.com/net-ssh/net-ssh-gateway/issues/4) which handily includes some example code to reporoduce the problem. 

I'm no expert but looking through the code, this line in **on_open_failed** [ in forwarder.rb](https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/service/forward.rb#L94) seems to cause the problem because i think it closed the underlying gateway channel which then stops all subsequent connections via the session from working. The session remains fine and is marked as active.

For me the problem occurs when i try to connect via the tunnel to a server with an invalid hostname.

If i comment out the line mentioned above then my script works ok.

thechile:
just some more info. I poked around some more and i think the gateway sessions transport is being closed.

So in the line i referenced originally, it calls `channel[:socket].close` and according to the documents `close` will wait for channels to close and then also "..closes the underlying transport layer connection"

Since the gateways transport is closing it brakes future use. I traced to the point wher it actually hangs after the transport is killed and it's in server_version.rb negotiate. It timeouts after 20 seconds on the `IO.select`, i assume because the transport is gone.

    if timeout && !IO.select([socket], nil, nil, timeout)
      raise Net::SSH::ConnectionTimeout, "timeout during server version negotiating"
    end

PR #286, opened 2015-11-27
Fix event loop lock ups when using net-ssh-gateway and multiple server connections

thechile:
As per https://github.com/net-ssh/net-ssh-gateway/issues/4 and https://github.com/net-ssh/net-ssh/issues/269 the PR fixes for me several problems I had when using persistent net-ssh-gateway sessions.

For some background info, I connect to several hundred servers located in various private/public clouds. These are only accessible via a jump server/bastion host. When I tried to open one net-ssh-gateway session to each of these gateway hosts and then connect onwards to the various servers.. my sessions would lock up and hang after a short amount of time. There wasn’t any real pattern or useful debug messages. It just seems to crash and burn.

After digging around and isolating the problem to the event loop it seems some channels were in strange states, or not being cleaned up correctly.

These changes allow me to connect to various jump servers via net-ssh-gateway and then connect to 100’s servers and run various commands without any problems.

### DIFF
--- a/lib/net/ssh/service/forward.rb
+++ b/lib/net/ssh/service/forward.rb
@@ -91,7 +91,10 @@ module Net; module SSH; module Service
 
         channel.on_open_failed do |ch, code, description|
           channel.error { "could not establish direct channel: #{description} (#{code})" }
-          channel[:socket].close
+          session.stop_listening_to(ch[:socket])
+          channel[:socket].close unless channel[:socket].closed?
+          channel.do_close
+          session.channels.delete(ch.local_id)
         end
       end
 
@@ -105,10 +108,12 @@ module Net; module SSH; module Service
     #   ssh.forward.cancel_local(1234)
     #   ssh.forward.cancel_local(1234, "0.0.0.0")
     def cancel_local(port, bind_address="127.0.0.1")
+      channel_id = session.lookup_channel_id_from_localport(port)
       socket = @local_forwarded_ports.delete([port, bind_address])
       socket.shutdown rescue nil
       socket.close rescue nil
       session.stop_listening_to(socket)
+      session.channels.delete(channel_id)
     end
 
     # Returns a list of all active locally forwarded ports. The returned value
--- a/lib/net/ssh/connection/channel.rb
+++ b/lib/net/ssh/connection/channel.rb
@@ -59,6 +59,9 @@ module Net; module SSH; module Connection
     # The remote id for this channel, assigned by the remote host.
     attr_reader :remote_id
 
+    # used to store the local port used by this channel
+    attr_accessor :local_port
+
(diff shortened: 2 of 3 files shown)


mfazekas:
Looks good. Is there any way we could reproduce the original issues, ideally we'd like to have tests to reproduce the issue.

thechile:
OK sure, i'll send some info ASAP on what i was doing to cause lockups.

lib/net/ssh/service/forward.rb
```diff
@@ -91,7 +91,10 @@ def local(*args)
 
         channel.on_open_failed do |ch, code, description|
           channel.error { "could not establish direct channel: #{description} (#{code})" }
-          channel[:socket].close
+          session.stop_listening_to(ch[:socket])
```
mfazekas: i'd write ```channel[:socket]``` instead of ```ch[:socket]``` for consistency

lib/net/ssh/service/forward.rb
```diff
@@ -91,7 +91,10 @@ def local(*args)
...
+          channel[:socket].close unless channel[:socket].closed?
+          channel.do_close
```
mfazekas: I'd question if we need to invoke ```on_close``` events in case ```on_open_failed```

mfazekas:
see also #325 for testcase

mfazekas:
actually i'm no even sure we should close that socket. It can be an intermittent failure and server can be started and further connects might suceed

lib/net/ssh/service/forward.rb
```diff
@@ -91,7 +91,10 @@ def local(*args)
...
+          session.channels.delete(ch.local_id)
```
mfazekas: https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/connection/session.rb#L561
we already delete channel, so this should not be neccesary

lib/net/ssh/service/forward.rb
```diff
@@ -105,10 +108,12 @@ def local(*args)
     #   ssh.forward.cancel_local(1234)
     #   ssh.forward.cancel_local(1234, "0.0.0.0")
     def cancel_local(port, bind_address="127.0.0.1")
+      channel_id = session.lookup_channel_id_from_localport(port)
       socket = @local_forwarded_ports.delete([port, bind_address])
       socket.shutdown rescue nil
       socket.close rescue nil
       session.stop_listening_to(socket)
+      session.channels.delete(channel_id)
```
mfazekas: i don't think this is neccesary such channels will be eventually closed and close will remove them from session. Also there can be multiple channels for a given port, so closing just the first doesn't sounds right

mfazekas:
please review #325 i think that is the correct fix, but let me know if you think otherwise

Issue #293, opened 2015-12-06
Minor UI issue when asking for a passphrase for encrypted key_data


Temikus:
When entering a passphrase for an encrypted ssh key stored as key_data, the UI states:
`Enter passphrase for :`
, which is somewhat ambiguous.

PR #294, opened 2015-12-06
Clarified prompt when key is supplied from data

Temikus:
Fixing #293 - minor UI change after my last change, to make the prompt:
`Enter passphrase for the encrypted key:`
Instead of:
`Enter passphrase for  :`

From what I see tests should be ok, since they define the prompt as: 
`reader.expect(/Enter passphrase for .*:/) { |data| puts data }`
, however, let's see what Travis reports.


mfazekas:
Clarified by #352

PR #296, opened 2015-12-10
Bugfix: proxy command was using nonblocking io api incorrectly causin…

### LINKED ISSUE #122: IO::EAGAINWaitReadable: Resource temporarily unavailable on jruby 1.7.4
Here is the environment
jruby 1.7.4

gem list
*** LOCAL GEMS ***

bouncy-castle-java (1.5.0146.1)
bundler (1.2.4)
jruby-launcher (1.0.15 java)
jruby-openssl (0.8.2)
net-ssh (2.7.0)
rake (10.0.3)
rubygems-bundler (1.1.0)
rvm (1.11.3.6)

When I try Net::SSH.start("xxx.xxxx.xxx", 'username') {|ssh| } i get this error
IO::EAGAINWaitReadable: Resource temporarily unavailable - errno backtraces disabled; run with -Xerrno.backtrace=true to enable

This might not be a net-ssh error, I am not sure.

mfazekas:
…g rare IO::EAGAIN errors

FIXES #122

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,4 @@
+* fixed rare WaitWritable error with proxy commands [Miklos Fazkas, Andre Meij]]
 * if Net::SSH.start user is nil and config has no entry we default to Etc.getlogin
 * Bugfix: CHANNEL_CLOSE was sent before draining ouput buffer #280 [Christopher F. Auston]
 
--- a/lib/net/ssh/proxy/command.rb
+++ b/lib/net/ssh/proxy/command.rb
@@ -79,11 +79,26 @@ module Net; module SSH; module Proxy
         end
       else
         def io.send(data, flag)
-          write_nonblock(data)
+          begin
+            result = write_nonblock(data)
+          rescue IO::WaitWritable, Errno::EINTR
+            IO.select(nil, [self])
+            retry
+          end
+          result
         end
 
         def io.recv(size)
-          read_nonblock(size)
+          begin
+            result = read_nonblock(size)
+          rescue IO::WaitReadable, Errno::EINTR
+            timeout_in_seconds = 20
+            if IO.select([self], nil, [self], timeout_in_seconds) == nil
+              raise "Unexpected spurious read wakeup"
+            end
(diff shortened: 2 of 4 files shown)


TiagoCardoso1983:
Shouldn't you be using the IO module from the project? 

https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/ruby_compat.rb

mfazekas:
Thanks, good catch.  Actually we should be remove/deprecate ruby_compat. As that was for 1.8 compat which we no longer support.

TiagoCardoso1983:
-1 on that. Having a separate IO module which proxies these calls is what allows me to patch them when using some other event loop like event machine or celluloid IO.

mfazekas:
@TiagoCardoso1983  We should have proper abstraction then, it was never meant for an extension point. Do you can code somewhere so i can see what you're doing exactly?

Note that i have an experimental, WIP EventLoop abstracted out, that would be much better point for doing such stuff:
https://github.com/mfazekas/net-ssh/blob/server_side/lib/net/ssh/connection/session.rb#L30

TiagoCardoso1983:
@mfazekas  not open source (yet), but here's the gist of it: https://gist.github.com/TiagoCardoso1983/ceaa809ff2bc9a461484 .

The thing about the loop is that one should have a way to switch between the dedicated loop (EM/Celluloid::IO) and the raw implementation within the same process. 

This EventLoop is quite recent, right? I'll have a look into it.

TiagoCardoso1983:
It would be cool if you could isolate the select call in its own method. This would make it extremely easy to just use inheritance and overwrite the select call. 

I don't know how to best approach this, since none of those other solutions provide an API quite like IO.select, all of them abstract the loop and make you act on the socket directly. Maybe net-ssh could do something of the kind and not multiplex on the different sessions/channels itself?

Here's an example of what I mean: https://github.com/brianmario/mysql2/blob/master/lib/mysql2/em.rb , which has an API for async send and async read. The wait is then done using the Deferable abstraction from event machine.

PR #297, opened 2015-12-19
Small DRY fix

MajesticFalcon:
First attempt at Git Hub.. When calling 'configuration_for' there is no reason that I can find to specify a default value for ' use_ssh_config' because Hash#fetch called on 'options' will always either return what the user specified or True. Just a small Do not repeat yourself.

### DIFF
--- a/lib/net/ssh.rb
+++ b/lib/net/ssh.rb
@@ -252,7 +252,7 @@ module Net
     # to read.
     #
     # See Net::SSH::Config for the full description of all supported options.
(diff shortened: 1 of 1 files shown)


mfazekas:
Thanks much, my only issue is that i'm not sure if anyone is calling that configuration_for directly. If so it might be an incompatible change. And the improvements doesn't seems to worth the risk to merge to 3.0. I'll probably make a branch for stuff like this which will end up in 4.0 where it's fine to have potentially non compat changes.

Issue #285, opened 2015-11-26
already initialized constant DL::RUBY_FREE


MarioRuiz:
when running net-ssh I got this message:

C:/Ruby21/lib/ruby/2.1.0/i386-mingw32/dl.so: warning: already initialized constant DL::RUBY_FREE
C:/Ruby21/lib/ruby/gems/2.1.0/gems/net-ssh-3.0.1/lib/net/ssh/authentication/pageant.rb:16: warning: previous definition of RUBY_FREE was here

it seems the problem is when using with mysql2 gem, if i run this:
require 'net/ssh'
require 'mysql2'
i got this message:
C:/Ruby21/lib/ruby/2.1.0/i386-mingw32/dl.so: warning: already initialized constant DL::RUBY_FREE
C:/Ruby21/lib/ruby/gems/2.1.0/gems/net-ssh-3.0.1/lib/net/ssh/authentication/pageant.rb:16: warning: previous definition of RUBY_FREE was here

if i run this:
require 'mysql2'
require 'net/ssh'
i got this:
C:/Ruby21/lib/ruby/gems/2.1.0/gems/net-ssh-3.0.1/lib/net/ssh/authentication/pageant.rb:15: warning: already initialized constant DL::CPtr
C:/Ruby21/lib/ruby/gems/2.1.0/gems/net-ssh-3.0.1/lib/net/ssh/authentication/pageant.rb:16: warning: already initialized constant DL::RUBY_FREE

PR #306, opened 2016-01-27
already initialized constant DL::RUBY_FREE #285

MarioRuiz:
To avoid this message when the constants have been already initialized by other gem


mfazekas:
thanks

Issue #307, opened 2016-02-03
KnownHosts caching?


byroot:
Hi, I'm working on Capistrano's performance (see https://github.com/capistrano/sshkit/issues/326 for more context), and it appear that `Net::SSH::KnownHosts#keys_for` is a big hot spot during connection establishment.

The profiled code:
```
require 'stackprof'
task :profile do
  StackProf.run(mode: :wall, out: 'tmp/stackprof-cap-idle-cold.dump') do
    on servers do # About ~180 hosts
      execute :echo, 'ping > /dev/null'
    end
  end
end
```

```
==================================
  Mode: wall(1000)
  Samples: 4981 (5.73% miss rate)
  GC: 394 (7.91%)
==================================
     TOTAL    (pct)     SAMPLES    (pct)     FRAME
      1880  (37.7%)        1874  (37.6%)     SSHKit::Runner::Parallel#execute
       476   (9.6%)         476   (9.6%)     block (3 levels) in Net::SSH::KnownHosts#keys_for
       859  (17.2%)         309   (6.2%)     block (2 levels) in Net::SSH::KnownHosts#keys_for
       276   (5.5%)         276   (5.5%)     Net::SSH::Compat.io_select
       183   (3.7%)         106   (2.1%)     OpenSSL::PKey::DH#valid?
       144   (2.9%)         102   (2.0%)     Net::SSH::Transport::Kex::DiffieHellmanGroup1SHA1#send_kexinit
        77   (1.5%)          77   (1.5%)     block in OpenSSL::PKey::DH#valid?
        63   (1.3%)          63   (1.3%)     block (2 levels) in Net::SSH::Transport::ServerVersion#negotiate!
       215   (4.3%)          53   (1.1%)     Net::SSH::Transport::PacketStream#poll_next_packet
        78   (1.6%)          52   (1.0%)     block in Net::SSH::Config.load
        50   (1.0%)          50   (1.0%)     Net::SSH::KnownHosts#known_host_hash?
        56   (1.1%)          47   (0.9%)     Net::SSH::Buffer#read
        46   (0.9%)          46   (0.9%)     block (2 levels) in Net::SSH::Connection::Channel#forward_local_env
       122   (2.4%)          44   (0.9%)     Net::SSH::Config.load
       934  (18.8%)          42   (0.8%)     Net::SSH::KnownHosts#keys_for
        70   (1.4%)          34   (0.7%)     Net::SSH::BufferedIo#fill
[...]

Issue #232, opened 2015-03-10
Separate/modularize known_hosts parsing


TiagoCardoso1983:
I'm using net-ssh in a concurrent scenario where I'm connecting to many hostnames in serial. This is making my known_hosts file (understandably) massively big, which makes the whole connection slow. But worse than that, the reading is happening x + y times, being x the number of hostnames I'm connecting to, and y the number of times that the known_hosts file is parsed after calling Net::SSH.start . My current known_hosts file sample has (gulp!) over 18000 entries. 

Now, I guess that maybe net-ssh wasn't thought for such scenarios, but I'm thinking whether one could easily modularize the whole known hosts file parsing, in that maybe the file would be held in memory (double gulp) and passed to the session initialization, maybe make it an additional parameter which is passed to the session initialization. I don't know, these are just suggestions and ideas. But currently as things stand, it's a bit unbearable, as parsing a > 18000-line file is causing me a huge number of EOF and broken pipes all around. Is there a short-term solution for it or should one discuss about possible solutions?

I assume that all ssh configuration files suffer from the same issue, that is, they are loaded during the session establishment flow.

mfazekas:
As a workaround you should be able to read the known hosts yourself then when connection generate a new temp file with just he needed entry and use :user_known_hosts_file to specify that host file.

carlhoerberg:
Or disable host checking completely with `paranoid: false`

TiagoCardoso1983:
@mfazekas I thought about that as a possible workaround, but, in the case where I will access all hostnames concurrently, the penalty will still hold. In the scenario I'm working with, that's a fairly doable scenario. 
@carlhoerberg wouldn't that defeat the purpose of using ssh in the first place? I'd be then susceptible to man-in-the-middle attacks, right? If I have to sacrifice security over performance, wouldn't I be then just ok with plain telnet?

carlhoerberg:
Well, with SSH your at risk of MITM on every first connect to a new server unless you've collected the host pub key beforehand through an out-of-band method, but maybe you're doing that?

But if you're using ssh key authentication (not password authentication) you still can't be MITMed as the MITM attacker (hopefully) don't have your private key, and the real server only accepts your private key. The MITM can still fake being a server and listen to the commands you're sending (which can be bad enough), but they won't reach the real server. 

Please correct me if I'm wrong.

TiagoCardoso1983:
[...]

PR #308, opened 2016-02-03
Allow to substitute Net::SSH::KnownHosts with a custom implementation

byroot:
Fix: https://github.com/net-ssh/net-ssh/issues/307
Fix: https://github.com/net-ssh/net-ssh/issues/232
Ref: https://github.com/capistrano/sshkit/issues/326

As suggested in https://github.com/net-ssh/net-ssh/issues/307#issuecomment-179345555.

I initially defaulted with `options[:known_hosts] = KnownHosts` in `Net::SSH.start`, but then I figured that some people might use only parts of `Net::SSH`, and so I think it's better to check the options in all the sub components.

@mfazekas @TiagoCardoso1983 what do you think of this implementation?

### DIFF
--- a/lib/net/ssh.rb
+++ b/lib/net/ssh.rb
@@ -66,7 +66,7 @@ module Net
       :keepalive, :keepalive_interval, :keepalive_maxcount, :kex, :keys, :key_data,
       :languages, :logger, :paranoid, :password, :port, :proxy,
       :rekey_blocks_limit,:rekey_limit, :rekey_packet_limit, :timeout, :verbose,
-      :global_known_hosts_file, :user_known_hosts_file, :host_key_alias,
+      :known_hosts, :global_known_hosts_file, :user_known_hosts_file, :host_key_alias,
       :host_name, :user, :properties, :passphrase, :keys_only, :max_pkt_size,
       :max_win_size, :send_env, :use_agent, :number_of_password_prompts,
       :append_supported_algorithms, :non_interactive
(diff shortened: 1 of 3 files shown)


TiagoCardoso1983:
I still think it's problematic to call the same method with the same arguments in two different parts of the path to session establishment. Under which circumstances are the arguments (if ever) going to be different? The known host key should be fetched as early as needed and stored somewhere for the second call to be optimized, what do you think?

mfazekas:
I like the change it's straightforward, doesn't breaks backward compatibility. @TiagoCardoso1983 i think you're after eliminating second ```KnownHosts.search_for```, i think that is ortogonal/independent to this change. From performance point of view doing a single search_for, and passing that down i assume in ```Session#initialize``` would be better, but that can be don in a different PR.

byroot:
> but that can be don in a different PR.

Indeed. I can try to fix it once this PR is merged.

mfazekas:
:+1: Thanks!

byroot:
So I looked into re-using the result of `search_for`, and it's not that easy because both calls are in quite different classes without references on each others.

Basically, their only sharing mechanism is the `options` hash. SO I think we could add yet another option `keys`, and if missing it would be populated with `KnownHosts.search_for` from `Session#initialize`.

@mfazekas how do you feel about that plan? I have to say I'm personally on the fence...

mfazekas:
@byroot i think we also have the option of passing via extra argument. I think options might be a bit overused. In case we use option we'd better use some less ambigous name like ```host_key_value``` 

```ruby
      host_key = known_hosts.search_for(options[:host_key_alias] || session.host_as_string, options)
      @host_key_verifier = select_host_key_verifier(options[:paranoid], host_key)
      @server_version = ServerVersion.new(socket, logger, options[:timeout])
      @algorithms = Algorithms.new(self, host_key, options)
```

PR #311, opened 2016-02-11
Memoize session host keys

byroot:
As discussed here: https://github.com/net-ssh/net-ssh/pull/308#issuecomment-179682869

`KnownHosts.search_for` always re-parse from zero and is called twice per session. This PR centralize and memoize that lookup.

It has been identified as a hot spot during profiling. See https://github.com/capistrano/sshkit/issues/326#issuecomment-178899236 for more details.

@mfazekas @TiagoCardoso1983 how does it look?

### DIFF
--- a/lib/net/ssh/transport/algorithms.rb
+++ b/lib/net/ssh/transport/algorithms.rb
@@ -242,8 +242,7 @@ module Net; module SSH; module Transport
           # make sure the host keys are specified in preference order, where any
           # existing known key for the host has preference.
 
-          known_hosts = options.fetch(:known_hosts, KnownHosts)
-          existing_keys = known_hosts.search_for(options[:host_key_alias] || session.host_as_string, options)
+          existing_keys = session.host_keys
           host_keys = existing_keys.map { |key| key.ssh_type }.uniq
           algorithms[:host_key].each do |name|
             host_keys << name unless host_keys.include?(name)
--- a/lib/net/ssh/transport/session.rb
+++ b/lib/net/ssh/transport/session.rb
@@ -89,6 +89,13 @@ module Net; module SSH; module Transport
       raise Net::SSH::ConnectionTimeout
     end
 
+    def host_keys
+      @host_keys ||= begin
+        known_hosts = options.fetch(:known_hosts, KnownHosts)
+        known_hosts.search_for(options[:host_key_alias] || host_as_string, options)
+      end
+    end
+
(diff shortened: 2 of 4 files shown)


TiagoCardoso1983:
@byroot looks good to me :+1:

mfazekas:
@byroot I like it in general.

There are some minor issue i'm thinking on:
 
```ruby
host = options[:host_key_alias] || arguments[:session].host_as_string
known_hosts = options.fetch(:known_hosts, KnownHosts)
host_keys = arguments[:session].host_keys
...
if host_keys.empty?
  known_host.add(host,args[:key],...)
...
```

On issue i'm worried about is that if someone would ask  ```arguments[:session].host_keys``` afterward it might return stale value/empty.

I'm thinking about adding ```add_host_key``` to ```host_keys``` so code would look like:

```ruby
host_keys = arguments[:session].host_keys
...
if host_keys.empty?
   host_keys.add_host_key(args[:key]...)
```

That would change requirement for :known_hosts to implement ```search_for``` and the returned enumerable should implement ```add_host_key(...)``` and ```host```

What do you think?

byroot:
That's not a bad idea, but that's quite a bigger API. The current `KnownHosts` would also need to be updated to that new API.

mfazekas:
It has to be decided as it will be painfull to change later for compatibility reasons. ```search_for``` could just return a subclass of array.
```ruby
class HostKeys < Array
  def initialize(hosts, host, known_hosts, options = {})
     super(hosts)
     @host = host
     @known_hosts = known_hosts
     @options = options
  end
  def add_host_key(key)
     @known_hosts.add(@host, key, options)
     push(key)
  end
end 
```

byroot:
I would advise against subclassing array, because then it exposes many methods that would bypass `add_host_key`.

mfazekas:
```ruby
class HostKeys
  include Enumerable
  attr_reader :host
  def initialize(host_keys, host, known_hosts, options = {})
     @host_keys = host_keys
     @host = host
     @known_hosts = known_hosts
     @options = options
  end
  def add_host_key(key)
     @known_hosts.add(@host, key, options)
     push(key)
  end
  def each
     @host_keys
  end
end
```
If we treat SemVer seriously we might need Array subclass as in theory there might be code there calling ```search_for``` - i don't think there is.

mfazekas:
https://github.com/net-ssh/net-ssh/compare/master...mfazekas:host-keys-cache?expand=1

PR #327, opened 2016-03-11
Fix Secure#verify call to process_cache_miss

byroot:
The regression was introduced in #320 

```
net/ssh/verifiers/secure.rb:41:in `process_cache_miss'
net/ssh/verifiers/secure.rb:33:in `verify'
net/ssh/verifiers/strict.rb:16:in `verify'
net/ssh/verifiers/lenient.rb:15:in `verify'
net/ssh/transport/kex/diffie_hellman_group1_sha1.rb:173:in `verify_server_key'
net/ssh/transport/kex/diffie_hellman_group1_sha1.rb:68:in `exchange_keys'
net/ssh/transport/algorithms.rb:363:in `exchange_keys'
net/ssh/transport/algorithms.rb:200:in `proceed!'
net/ssh/transport/algorithms.rb:191:in `send_kexinit'
net/ssh/transport/algorithms.rb:146:in `accept_kexinit'
net/ssh/transport/session.rb:209:in `block in poll_message'
net/ssh/transport/session.rb:187:in `loop'
net/ssh/transport/session.rb:187:in `poll_message'
net/ssh/transport/session.rb:224:in `block in wait'
net/ssh/transport/session.rb:222:in `loop'
net/ssh/transport/session.rb:222:in `wait'
net/ssh/transport/session.rb:87:in `initialize'
net/ssh.rb:228:in `new'
net/ssh.rb:228:in `start'
```

@mfazekas

PR #329, opened 2016-03-14
Fix KnownHosts#add reference to options

byroot:
The regression was introduced in #320
Ref: https://github.com/net-ssh/net-ssh/pull/327

@mfazekas I missed that one until now because so far I had all the keys already added.


mfazekas:
Thanks much 3.1.0rc2 is released now with this fix

Issue #316, opened 2016-03-05
/net/ssh/authentication/key_manager.rb:65:in `expand_path' - no implicit conversion of nil into String


Startouf:
When trying to establish a SSH connection for deployment with Capistrano, I had a problem loading the private key file for puclic key authentication, and the only message I had was a 

```
TypeError: no implicit conversion of nil into String
C:/Dev/Ruby22-x64/lib/ruby/gems/2.2.0/gems/net-ssh-3.0.2/lib/net/ssh/authentication/key_manager.rb:65:in `expand_path'
...
```

After digging net-ssh code and debugging my own code I realized it was just because there was a mistype in my configuration and I was providing a nil parameter for the private key file. Would it be okay to check that the given parameter is indeed a valid string pointing to a valid private key file ?

mfazekas:
@Startouf so your usecase is passing ```keys: [ENV['foo']] ``` where foo is not defined?

Startouf:
Yes, which resulted most likely in `[nil]` being passed.

mfazekas:
@Startouf sorry for late answer yes if you can add a check that would be great. i think the best place would be  where we extract from options. https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/authentication/session.rb#L146

PR #355, opened 2016-04-17
Prevents users explicitly passing in nils in options by accident whic…

KaneMorgan:
…h could lead to no method errors further down

I think this fixes https://github.com/net-ssh/net-ssh/issues/316

### DIFF
--- a/lib/net/ssh.rb
+++ b/lib/net/ssh.rb
@@ -178,7 +178,7 @@ module Net
     # * :user_known_hosts_file => the location of the user known hosts file.
     #   Set to an array to specify multiple user known hosts files.
     #   Defaults to %w(~/.ssh/known_hosts ~/.ssh/known_hosts2).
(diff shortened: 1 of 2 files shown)


mfazekas:
@KaneMorgan looks good is there any option where nil might make sense?

KaneMorgan:
I don't know the library that well, but from what I could see setting it to nil explicitly will stop it using defaults which will normally just make it blow up. So in that case its probably better to fail earlier. 

The only use case I can think of is if you wanted to use nil's falseyness to set the value of a boolean. it'd be kind of a strange way to do it though.

From the VALID_OPTIONS i think maybe used with keys_only, password_prompt and non_interactive. look like booleans. Do you want me to allow those through?

mfazekas:
it's fine it's going to go to 4.0.0, we can wait until we have complaints about nil used as boolean.

PR #360, opened 2016-04-27
Added ability to specify agent socket

alongoldboim:
Adding the option to specify a socket that will be used, this is useful in case we have multiple agents running and multiple users using the net-ssh at the same time (that's why changing the ENV variable is not really practical).

### DIFF
--- a/lib/net/ssh/service/forward.rb
+++ b/lib/net/ssh/service/forward.rb
@@ -357,7 +357,7 @@ module Net; module SSH; module Service
         channel[:invisible] = true
 
         begin
-          agent = Authentication::Agent.connect(logger)
+          agent = Authentication::Agent.connect(logger, session.options[:agent_socket_factory])
           if (agent.socket.is_a? ::IO)
             prepare_client(agent.socket, channel, :agent)
           else
--- a/lib/net/ssh.rb
+++ b/lib/net/ssh.rb
@@ -71,7 +71,7 @@ module Net
       :known_hosts, :global_known_hosts_file, :user_known_hosts_file, :host_key_alias,
       :host_name, :user, :properties, :passphrase, :keys_only, :max_pkt_size,
       :max_win_size, :send_env, :use_agent, :number_of_password_prompts,
-      :append_supported_algorithms, :non_interactive, :password_prompt
+      :append_supported_algorithms, :non_interactive, :password_prompt, :agent_socket_factory
     ]
 
     # The standard means of starting a new SSH connection. When used with a
@@ -195,6 +195,9 @@ module Net
     #   to prefer failing a password/etc auth methods vs asking for password
     # * :password_prompt => a custom prompt object with ask method. See Net::SSH::Prompt
     #
+    # * :agent_socket_factory => enables the user to pass a lambda/block that will serve as the socket factory
+    #    Net::SSH::start(user,host,agent_socket_factory: ->{ UNIXSocket.open('/foo/bar') })
+    #    example: ->{ UNIXSocket.open('/foo/bar')}
     # If +user+ parameter is nil it defaults to USER from ssh_config, or
     # local username
     def self.start(host, user=nil, options={}, &block)
--- a/lib/net/ssh/authentication/agent/java_pageant.rb
+++ b/lib/net/ssh/authentication/agent/java_pageant.rb
@@ -19,7 +19,7 @@ module Net; module SSH; module Authentication
 
     # Instantiates a new agent object, connects to a running SSH agent,
     # negotiates the agent protocol version, and returns the agent object.
-    def self.connect(logger=nil)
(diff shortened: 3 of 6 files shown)


mfazekas:
Thanks much for the PR, can you please describe your usecase?
Rather than passing a connection have you considered passing a factory that makes the connection?

alongoldboim:
@mfazekas didn't quite get your comment, could you elaborate please?

mfazekas:
Sorry i was not reading your change carefully enough. I assumed `agent_socket` is an actual socket, but since you're passing it to ```agent_socket_factory.open``` it's likely meant to be an address rather than socket.

So your version is this:
```ruby
Net::SSH::start(user,host,agent_socket_addres:'/foo/bar')
```

what i was asking is this:
```ruby
Net::SSH::start(user,host,agent_socket_factory: ->{ UNIXSocket.open('/foo/bar') })
```

But again, please explain your usecase.

alongoldboim:
> Adding the option to specify a socket that will be used, this is useful in case we have multiple agents running and multiple users using the net-ssh at the same time (that's why changing the ENV variable is not really practical).
>  :)

@mfazekas thanks for the review, added needed fixes.

mfazekas:
multiple users with multiple agents sounds an esoteric usecase.

I find ```custom_socket_factory```  name confusing. We only use it for agent connections, so the name would be better to reflect that.

alongoldboim:
its a multi tenant app that uses net-ssh for some actions, the user inserts his private key, hence all the password will  be cached on the same ssh-agent which is a security issue.

the name you suggested 'agent_socket_factory' is the name of the existing function, so i can't use it, what other name do you suggest?

simon3z:
> multiple users with multiple agents sounds an esoteric usecase.

Hi @mfazekas think of a multi-tenant management application.
You want different users to be able to manage hosts belonging to their tenant... but you wouldn't want to collect all the keys in the same ssh-agent.

And remember that "multiple users with multiple agents" is the original design of ssh-agent. In a linux machine you run an ssh-agent per user, because ssh-agent is not multi-tenant itself and you don't want to cache all the keys in a single place for all the users.

mfazekas:
I'd suggest renamig that method it can be just ```socket_class```, and have the option named as ```agent_socket_factory``` or ```agent_connection_factory```.

mfazekas:
@simon3z, @alongoldboim thanks for clarifying the usecase, it makes sense that way.

alongoldboim:
@mfazekas Needed changes were made and tests fixed, please review again :)

lib/net/ssh/service/forward.rb
```diff
@@ -357,7 +357,7 @@ def auth_agent_channel(session, channel, packet)
         channel[:invisible] = true
 
         begin
-          agent = Authentication::Agent.connect(logger)
+          agent = Authentication::Agent.connect(logger, self.session.options[:agent_socket_factory])
```
mfazekas: can't we just write ```session``` vs ```self.session```

mfazekas:
Looks good to me, a test would be good to have.

lib/net/ssh/service/forward.rb
alongoldboim: fixed.

alongoldboim:
@mfazekas Added test to make sure that if we pass a lambda it will use its value instead of the regular ENV.

mfazekas:
Thanks much!

simon3z:
@mfazekas thanks!

@mfazekas is there any chance to have a new release soon?

mfazekas:
i'll try to release 4.0.0.alpha4 this weekend

alongoldboim:
Hey @mfazekas, I'm trying the new fix and i got a small problem i though you might help me find the reason.
```ruby 
require 'net/ssh'
agent_socket = "/tmp/ssh/ssh_a"
FileUtils.mkdir_p '/tmp/ssh_manageiq'
system "ssh-agent -a #{agent_socket}"
Net::SSH.start('10.35.4.213', 'root', :paranoid => false, :forward_agent => true,:agent_socket_factory => ->{ UNIXSocket.open(agent_socket) },
               :key_data => ['key']) do |ssh|
  res = ssh.exec!("ssh -o 'StrictHostKeyChecking no' root@10.35.4.163" + " echo $?")
end
```

alongoldboim:
@mfazekas i though net-ssh will add the key we connect with to the running agent automatically, apparently it doesn't, ill just do it manually, maybe we should consider adding the current key to the agent by default?

mfazekas:
I assume there might be usecase where auto adding the programatically defined key_data to agent automatically is not desirable. So maybe an option to add that to session.

alongoldboim:
@mfazekas your right, when ill get some time ill do a patch for that as well.
Regarding the release any ETA on that?

Issue #361, opened 2016-05-01
net-ssh doesn't work with pageant on cygwin


duerst:
This is the reverse of issue https://github.com/net-ssh/net-ssh/issues/192, in that I work on cygwin (linux emulation on Windows), but would like to use PuTTY pageant as an authentication agent. I get it to work if I comment out the checks around lines 4 and 136 in net/ssh/authentication/agent/socket.rb, but that's just a short-time quick hack.

Both issues could be solved by introducing an optional configuration variable that tells net-ssh to use (bash) ssh-agent or (PuTTY) pageant. If that helps, I can send a patch.

mfazekas:
@duerst #360 introduces agent_socket_factory i assume that can be used to solve it.

duerst:
@mfazekas: Thanks for the pointer. Line 136 that I mentioned in net/ssh/authentication/agent/socket.rb is indeed in agent_socket_factory. It's just that that line, which reads
    if Net::SSH::Authentication::PLATFORM == :win32
is too simplistic for my case or that of  issue #192.

What would the best way be to make this more flexible? The easiest thing I could come up with would be to allow Net::SSH::Authentication::PLATFORM to be defined in advance, e.g. in Capfile or wherever the user wants, as follows:
module Net; module SSH; module Authentication; PLATFORM = :win32; end end end
(A simple Net::SSH::Authentication::PLATFORM = :win32 doesn't work because the modules don't exist yet.)

As far as I understand, the only thing needed to make this work would be to change line 6 in 
from
  PLATFORM = File::ALT_SEPARATOR \
    ? RUBY_PLATFORM =~ /java/ ? :java_win32 : :win32 \
    : RUBY_PLATFORM =~ /java/ ? :java : :unix
to
  PLATFORM ||= File::ALT_SEPARATOR \
    ? RUBY_PLATFORM =~ /java/ ? :java_win32 : :win32 \
    : RUBY_PLATFORM =~ /java/ ? :java : :unix
to give preference to already existing values.

I have created a tentative pull request for this change at https://github.com/net-ssh/net-ssh/pull/362, just so that it's easy to see what change I propose. But something like
module Net; module SSH; module Authentication; PLATFORM = :win32; end end end
doesn't feel like Ruby; much more like a bad hack. Any advice welcome.

mfazekas:
So #360 introduces a new option to use like this:
[...]

PR #362, opened 2016-05-02
Tentative quick fix proposal for issue #361

duerst:
This is a tentative quick fix proposal as discussed in issue #361. But I don't feel this is the best solution; it should be easier for the user to indicate what's wanted.


mfazekas:
Thans for the contribution this was merged with slight modifications into master

Issue #192, opened 2014-10-22
net-ssh doesn't work with ssh-agent/requires Pageant on Windows


T2L:
### Issue summary

I'm using Capistrano on Windows 7.

I've configured ssh-agent to autoload  with bash, enabled agent forwarding in ssh config file. Server allows for agent forwarding as well.

Agent forwarding works as intended with `ssh user@exmaple.com` in bash.

However, it doesn't work with Capistrano, which uses net-ssh. 

I'm not Ruby developer, but I was able to debug it (using Logger:DEBUG): 

```
    servers: ["example.com"]
    [example.com] executing command
E, [2014-10-22T12:27:07.137433 #5852] ERROR -- net.ssh.authentication.agent[1a15688]: could not connect to ssh-agent
E, [2014-10-22T12:27:07.137433 #5852] ERROR -- net.ssh.service.forward[1a59938]: attempted to connect to agent but failed: Net::SSH::Authentication::AgentNotAvailable (pageant process not running)
E, [2014-10-22T12:27:07.137433 #5852] ERROR -- net.ssh.connection.session[1a4e4d8]: [2, "could not connect to authentication agent"]
 ** [example.com :: out] Error reading response length from authentication socket.
```

As far as I can tell, net-ssh tries to find Pageant process and fails on this. I **do not have** PuTTY, because I'm using bash (coming with Git)

I've found some piece of code that tries to open socket using OpenSSH agent (https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/authentication/agent/socket.rb#L65): 

``` ruby
@socket = agent_socket_factory.open(ENV['SSH_AUTH_SOCK'])
```

which seems like is trying to connect to ssh-agent process, but somehow this action fails.

So, I was forced to install PuTTY and run Pageant in order to get capistrano work.

Issues I see:
- State in documentation that net-ssh requires Pageant process in order to agent forwarding work (and it doesn't work with ssh-agent)
- If it does work with ssh-agent, fix integration
### Environment details
- Windows 7
- ruby 2.1.3p242 (2014-09-19 revision 47630) [i386-mingw32]
- net-ssh 2.9.1
- `ruby -ropenssl -e puts OpenSSL::OPENSSL_VERSION'` prints `OpenSSL 1.0.0n 6 Aug 2014`
- bash `openssl version` prints `OpenSSL 0.9.8zb 6 Aug 2014`
[...]

PR #365, opened 2016-05-04
Improve automatic selection of pageant on cygwin and win32

duerst:
The logic for selecting an agent on Windows (including cygwin) is now
as follows:
- if ENV[SSH_AUTH_SOCK] is defined, use UNIXSocket
- otherwise, use Pageant::Socket

This addresses both https://github.com/net-ssh/net-ssh/issues/192 and
https://github.com/net-ssh/net-ssh/issues/361.

### DIFF
--- a/lib/net/ssh/authentication/agent.rb
+++ b/lib/net/ssh/authentication/agent.rb
@@ -3,9 +3,11 @@ require 'net/ssh/errors'
 require 'net/ssh/loggable'
 
 module Net; module SSH; module Authentication
-  PLATFORM = File::ALT_SEPARATOR \
+  PLATFORM = RUBY_PLATFORM =~ /cygwin/ ? :win32 : # :win32 includes cygwin
+    (File::ALT_SEPARATOR \
     ? RUBY_PLATFORM =~ /java/ ? :java_win32 : :win32 \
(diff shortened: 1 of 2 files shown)


duerst:
@mfazekas: Ping!

lib/net/ssh/authentication/agent.rb
```diff
@@ -3,9 +3,11 @@
 require 'net/ssh/loggable'
 
 module Net; module SSH; module Authentication
-  PLATFORM = File::ALT_SEPARATOR \
+  PLATFORM = RUBY_PLATFORM =~ /cygwin/ ? :win32 : # :win32 includes cygwin
+    (File::ALT_SEPARATOR \
     ? RUBY_PLATFORM =~ /java/ ? :java_win32 : :win32 \
-    : RUBY_PLATFORM =~ /java/ ? :java : :unix
+    : RUBY_PLATFORM =~ /java/ ? :java : :unix)
+  PLATFORM = :cygwin if RUBY_PLATFORM =~ /cygwin/
```
mfazekas: In Line 6 we set PLATFORM to ```:win32``` on cygwin and in line 10 we set ```:cygwin``` why?

mfazekas:
@duerst can you check 4.0.0.rc1? It should have this logic implemented. Thanks again for the patch!

PR #367, opened 2016-05-05
Adds minimum_dh_bits option.

abedra:
This commit introduces a new option, minimum_dh_bits. This option allows
the user to specify the minimum required bits for a diffie helman key
exchange in situations where the minimum hardcoded value of 1024 is too
weak.


mfazekas:
@abedra, thanks looks good. For reference can you explain the usecase? I assume you want to prevent connecting to a server with not strong enough dh?!

abedra:
That's correct.

Issue #357, opened 2016-04-21
4.0.0.alpha3 windows 10: could not open library '.dll'


lumean:
Hi,
On win10 I get the following error when trying to load net-ssh 4.0.0.alpha3

```
 (LoadError)223-p173/lib/ruby/gems/2.2.0/gems/ffi-1.9.10-x86-mingw32/lib/ffi/library.rb:133:in `block in ffi_lib': Could not open library '.dll': Das angegebene Modul wurde nicht gefunden.
        from c:/ruby/ruby223-p173/lib/ruby/gems/2.2.0/gems/ffi-1.9.10-x86-mingw32/lib/ffi/library.rb:100:in `map'
        from c:/ruby/ruby223-p173/lib/ruby/gems/2.2.0/gems/ffi-1.9.10-x86-mingw32/lib/ffi/library.rb:100:in `ffi_lib'
        from c:/ruby/ruby223-p173/lib/ruby/gems/2.2.0/gems/rbnacl-3.3.0/lib/rbnacl/sodium.rb:10:in `extended'
        from c:/ruby/ruby223-p173/lib/ruby/gems/2.2.0/gems/rbnacl-3.3.0/lib/rbnacl/sodium/version.rb:9:in `extend'
        from c:/ruby/ruby223-p173/lib/ruby/gems/2.2.0/gems/rbnacl-3.3.0/lib/rbnacl/sodium/version.rb:9:in `<module:Version>'
        from c:/ruby/ruby223-p173/lib/ruby/gems/2.2.0/gems/rbnacl-3.3.0/lib/rbnacl/sodium/version.rb:6:in `<module:Sodium>'
        from c:/ruby/ruby223-p173/lib/ruby/gems/2.2.0/gems/rbnacl-3.3.0/lib/rbnacl/sodium/version.rb:4:in `<module:RbNaCl>'
        from c:/ruby/ruby223-p173/lib/ruby/gems/2.2.0/gems/rbnacl-3.3.0/lib/rbnacl/sodium/version.rb:3:in `<top (required)>'
        from c:/ruby/ruby223-p173/lib/ruby/site_ruby/2.2.0/rubygems/core_ext/kernel_require.rb:55:in `require'
        from c:/ruby/ruby223-p173/lib/ruby/site_ruby/2.2.0/rubygems/core_ext/kernel_require.rb:55:in `require'
[...]

PR #371, opened 2016-05-15
set default options before checking for nils

KaneMorgan:
Allows it to handle logger being set explicitly to nil which may happen if options are wrapped by another program. Some discussion in #357


mfazekas:
👍  thanks much!

PR #378, opened 2016-05-19
prevent runaway zombies

p-sherratt:
detach processes spawned by Net::SSH::Proxy::Command to prevent them turning into zombies

### DIFF
--- a/lib/net/ssh/proxy/command.rb
+++ b/lib/net/ssh/proxy/command.rb
@@ -56,6 +56,7 @@ module Net; module SSH; module Proxy
       }
       begin
         io = IO.popen(command_line, "r+")
+        Process.detach(io.pid)
         if result = Net::SSH::Compat.io_select([io], nil, [io], 60)
           if result.last.any?
             raise "command failed"


mfazekas:
Ideally we should be calling ```.close``` in the object returned by the Proxy. That should take care too.

Do you have a concrete usecase where net-ssh leaves around proxy processes?

mfazekas:
i'd really like to understand why we have those zombies so closing until we know more about that

willbryant:
Because that is how unix works.  Even when a process exits, it stays in the operating system's process table so that the parent process can find out what its exit code was.  It continues to therefore take up certain resources.

If you are not interested in the process's exit code, you must "detach" it, which means that the init daemon (process 1 - eg. systemd) will "wait" for it instead of your program having to do it.

mfazekas:
@willbryant that is not how ruby works. The close on IO object returned by popen will call waitpid. If you use net-ssh correctly closing the session it will close the proxy IO object that will call waitpid. No need for Process.detach unless you're misusing the library or there is a bug in the library.
In any case Process.detach is not the right fix for the issue

willbryant:
Hmm yeah OK, sorry.  It could be from the net-ssh-gateway gem.

PR #384, opened 2016-06-10
Fix https://github.com/net-ssh/net-ssh/issues/235

elconas:
This fixes the sporadic on our ruby: 

- ruby  2.2.4p230 (2015-12-16 revision 53155) [x64-mingw32]

```
the bignum too big to convert into `long'
```

issue


elconas:
Rebased in #385

PR #390, opened 2016-06-20
Changing the syntax of singleton methods in CTR cipher

ckaushik:
Fixing defect #274

defining singleton methods using `class << obj... def` is not threadsafe. Using an alternative syntax to work around this issue


mfazekas:
Fixes: #274

PR #394, opened 2016-07-15
Prevent encoding issues building UTF8 buffers (4.0 alpha branch)

Iristyle:
- Prior to this change, attempting to send UTF8 commands through
  SSH, or attempting to copy files with UTF8 filenames could fail.
  This was particularly easy to trigger by attempting to execute
  commands that were 128 bytes or longer.

- monkey patch net-ssh gem to allow UTF-8 strings >= 128 bytes

  The buffer @content is often built as a UTF-8 string, until the
  point at which it appends data that cannot be encoded as a UTF-8
  sequence.

  One case occurs when the call to write_string is made to append a
  string that exceeds 127 bytes in length.  The SSH2 format says
  that strings must be length prefixed, and when the value [128]
  has pack("N*") called against it, the resultant 4 byte network
  order representation does not have a valid UTF-8 equivalent,
  resulting in an ASCII-8BIT / BINARY string.

  [127].pack('N*').encode('utf-8')
  => "\u0000\u0000\u0000\u007F"

  [128].pack('N*').encode('utf-8')
  Encoding::UndefinedConversionError: "\x80" from ASCII-8BIT to UTF-8

  Ruby has a subtle behavior where appending a BINARY string to
  an existing UTF-8 string is allowed and the resultant string
  changes encoding to BINARY.  However, once this has happened,
  the string can no longer have UTF-8 encoded strings appended as
  Ruby will raise an Encoding:CompatibilityError

  Appending BINARY to UTF-8 always creates BINARY:
  "foo".encode('utf-8') << [128].pack('N*')
  => "foo\x00\x00\x00\x80"

  Appending UTF-8 representable strings to existing strings:

  Ruby 2.1.7 keeps the string as its default UTF-8
  "foo" << [127].pack('N*')
  => "foo\u0000\u0000\u0000\u007F"

  Ruby 1.9.3 keeps UTF-8 strings as UTF-8
  "foo".encode('utf-8') << [127].pack('N*')
  => "foo\u0000\u0000\u0000\u007F"

  Ruby 1.9.3 defaults to US-ASCII which changes it to BINARY
  pry(main)> "foo" << [127].pack('N*')
  => "foo\x00\x00\x00\x7F"

  The simple solution is to call force_encoding on UTF-8 strings
  prior to appending them to @content, given it's always OK to
  append ASCII-8BIT / BINARY strings to existing strings, but
  appending UTF-8 to BINARY raises errors.

  "\x80".force_encoding('ASCII-8BIT') << "\u16A0"
  Encoding::CompatibilityError: incompatible character encodings: ASCII-8BIT and UTF-8

  force_encoding in this case, will simply translate a valid UTF-8
  string to its BINARY equivalent

  "\u16A0".force_encoding('BINARY')
  => "\xE1\x9A\xA0"
  Correct conversion per http://www.fileformat.info/info/unicode/char/16a0/index.htm

### DIFF
--- a/test/test_buffer.rb
+++ b/test/test_buffer.rb
@@ -26,6 +26,21 @@ class TestBuffer < NetSSHTest
     assert_equal "\1\0\0\0\2\0\0\0\0\0\0\0\3\0\0\0\0014\1\0\000\000\000\004I\226\002\322something", buffer.to_s
   end
 
+  def test_from_should_build_new_buffer_that_includes_utf8_string_128_characters
+    length = 128
+    # letter A has a 1 byte UTF8 representation
+    buffer = Net::SSH::Buffer.from(:long, 2, :string, 'A' * length)
+    # long of 2 + length 128 as network endian + 128 A's
+    expected = "\x00\x00\x00\x02" + [length].pack('N*') + ("\x41" * length)
+    assert_equal expected, buffer.to_s
+  end
+
+  def test_from_should_build_new_buffer_with_frozen_strings
+    foo = 'foo'.freeze
+    buffer = Net::SSH::Buffer.from(:string, foo)
+    assert_equal "\0\0\0\3foo", buffer.to_s
+  end
+
   def test_from_with_array_argument_should_write_multiple_of_the_given_type
     buffer = Net::SSH::Buffer.from(:byte, [1,2,3,4,5])
     assert_equal "\1\2\3\4\5", buffer.to_s
@@ -33,7 +48,7 @@ class TestBuffer < NetSSHTest
 
   def test_from_should_measure_bytesize_of_utf_8_string_correctly
     buffer = Net::SSH::Buffer.from(:string, "\u2603") # Snowman is 3 bytes
-    assert_equal "\0\0\0\3\u2603", buffer.to_s
+    assert_equal "\0\0\0\3\u2603".force_encoding('BINARY'), buffer.to_s
   end
 
   def test_read_without_argument_should_read_to_end
--- a/lib/net/ssh/buffer.rb
+++ b/lib/net/ssh/buffer.rb
@@ -182,7 +182,7 @@ module Net; module SSH
       consume!
       data
     end
-      
+
     # Return the next 8 bytes as a 64-bit integer (in network byte order).
     # Returns nil if there are less than 8 bytes remaining to be read in the
     # buffer.
@@ -284,7 +284,7 @@ module Net; module SSH
     # Writes the given data literally into the string. Does not alter the
     # read position. Returns the buffer object.
     def write(*data)
-      data.each { |datum| @content << datum }
+      data.each { |datum| @content << (datum.frozen? ? datum.encode('BINARY') : datum.force_encoding('BINARY')) }
       self
     end
 


Iristyle:
This targets the next 4.0.0.alpha4 release.

FYI - I have opened 4 PRs to address this problem:

* #394 - 4.x alpha branch - 4.0.0.alpha4 gem last cut 5/2/2016
* #395 - 2.9 branch - 2.9.4 gem last cut 1/30/2016
* #396 - 2.10 branch - 2.10.1.rc2 gem last cut 9/8/2015
* #397 - 3.1 branch - 3.2.0 gem last cut 6/19/2016

I couldn't quite figure out how your branches are setup, so I targeted all recent releases.

test/test_buffer.rb
```diff
@@ -26,14 +26,23 @@ def test_from_should_build_new_buffer_from_definition
...
+  end
+
   def test_from_with_array_argument_should_write_multiple_of_the_given_type
     buffer = Net::SSH::Buffer.from(:byte, [1,2,3,4,5])
     assert_equal "\1\2\3\4\5", buffer.to_s
   end
 
   def test_from_should_measure_bytesize_of_utf_8_string_correctly
     buffer = Net::SSH::Buffer.from(:string, "\u2603") # Snowman is 3 bytes
-    assert_equal "\0\0\0\3\u2603", buffer.to_s
```
mfazekas: I'd prefer ```assert_equal  "\0\0\0\3\u2603".force_encoding('BINARY'), buffer.to_s```

mfazekas:
@Iristyle thanks much the patch looks great! 
We'd only need to fix that rubocop error - see https://travis-ci.org/net-ssh/net-ssh/jobs/145133832#L418

I'd prefer the change to go to master first, then we can backport to ```3.2.*``` i wouldn't deal with 2.9 unless someone has a compelling argument.

👍

mfazekas:
@Iristyle there is also an issue with the following sample:

```ruby
require 'net/ssh'

username = ENV['USER']
host = 'localhost'
ssh = Net::SSH.start(host, username)
puts ssh.exec!('pwd')
```

output:
```ruby
/Work/Freelance/Ruby/NetSSH/net-ssh/lib/net/ssh/buffer.rb:287:in `force_encoding': can't modify frozen String (RuntimeError)
	from /Work/Freelance/Ruby/NetSSH/net-ssh/lib/net/ssh/buffer.rb:287:in `block in write'
	from /Work/Freelance/Ruby/NetSSH/net-ssh/lib/net/ssh/buffer.rb:287:in `each'
	from /Work/Freelance/Ruby/NetSSH/net-ssh/lib/net/ssh/buffer.rb:287:in `write'
	from /Work/Freelance/Ruby/NetSSH/net-ssh/lib/net/ssh/buffer.rb:326:in `block in write_string'
	from /Work/Freelance/Ruby/NetSSH/net-ssh/lib/net/ssh/buffer.rb:323:in `each'
	from /Work/Freelance/Ruby/NetSSH/net-ssh/lib/net/ssh/buffer.rb:323:in `write_string'
	from /Work/Freelance/Ruby/NetSSH/net-ssh/lib/net/ssh/buffer.rb:55:in `block in from'
	from /Work/Freelance/Ruby/NetSSH/net-ssh/lib/net/ssh/buffer.rb:47:in `step'
	from /Work/Freelance/Ruby/NetSSH/net-ssh/lib/net/ssh/buffer.rb:47:in `from'
	from /Work/Freelance/Ruby/NetSSH/net-ssh/lib/net/ssh/authentication/methods/abstract.rb:45:in `userauth_request'
	from /Work/Freelance/Ruby/NetSSH/net-ssh/lib/net/ssh/authentication/methods/none.rb:13:in `authenticate'
	from /Work/Freelance/Ruby/NetSSH/net-ssh/lib/net/ssh/authentication/session.rb:80:in `block in authenticate'
	from /Work/Freelance/Ruby/NetSSH/net-ssh/lib/net/ssh/authentication/session.rb:66:in `each'
	from /Work/Freelance/Ruby/NetSSH/net-ssh/lib/net/ssh/authentication/session.rb:66:in `authenticate'
	from /Work/Freelance/Ruby/NetSSH/net-ssh/lib/net/ssh.rb:241:in `start'
	from ./tryout/test_force_encoding.rb:9:in `<main>'
```

Iristyle:
Ah yeah, that's because `ENV['USER']` is frozen.  Good catch.

The advantage of `force_encoding` is that you don't create another copy of the string - but when  a string is frozen, you have to do that to change it.

So you have a couple of choices here:

* `.encode('BINARY')` instead of `.force_encoding` - which incurs a string copy
* `datum.frozen? ? datum.encode('BINARY') : datum.force_encoding('BINARY')` - mutates the string if not frozen, otherwise copies it

Iristyle:
Also - I think our Beaker library (at Puppet) is still pointed at `net-ssh` 2.9, which is why I targeted a PR there as well - see the gemspec
https://github.com/puppetlabs/beaker/blob/master/beaker.gemspec#L40

I'm not the primary maintainer of Beaker though, so I'm not sure if there's a reason that my teammates haven't looked at upgrading yet - @kevpl / @tvpartytonight might be able to comment further on that.

mfazekas:
I like ```datum.frozen? ? datum.encode('BINARY') : datum.force_encoding('BINARY')``` 

my only concern is this will still change parameters passed down to net-ssh if they are non frozen:
```Net::SSH.start(host, username)```
should the program reuse those objects other places, then it can have surprising consequences.

BTW ```encode('BINARY')``` seems to break for non ascii use:

```ruby
'é'.encode('BINARY')
Encoding::UndefinedConversionError: U+00E9 from UTF-8 to ASCII-8BIT
```

 ``` 'é'.dup.force_encoding('BINARY') ```

might be an alternative.

Iristyle:
Yes, `force_encoding` does mutate (but it doesn't mutate contents of the string, just the metadata for the string).  It's possible that this may cause strings to render in unexpected ways depending on usage.  It will be safer to make copies, though heavier to create dupe strings.

That conversion error you've highlighted is rather peculiar (I've found Ruby to be pretty poor when it comes to handling encodings in general).  There shouldn't be any reason that you can't convert that UTF-8 codepoint to a series of bytes (all UTF-8 and UTF-16 codepoints have 1 - 4 byte representations).

```
[1] pry(main)> "\u00E9".force_encoding('BINARY')
=> "\xC3\xA9"
```

http://www.fileformat.info/info/unicode/char/00e9/index.htm

mfazekas:
I'm thinking on defaulting to ```datum.dup.force_encoding('BINARY')``` then where we know it's we don't mind mutating the object then we'd call datum.force_encoding('BINARY').

I guess the encode("BINARY") issue cames because BINARY is just an alias to ASCII-8BIT, and it's reasonable that not every utf8 string can be transcoded to ascii.
```ruby
Encoding.aliases["BINARY"]
 => "ASCII-8BIT"
```

Iristyle:
@mfazekas it is correct that `ASCII-8BIT` and `BINARY` are synonyms for one another, but the `ASCII` name there is poorly chosen IMHO (it really doesn't have anything to do with the limits imposed by the `ASCII` charset).

Any UTF-8 string can be treated as a `BINARY` string - it simply means that instead of using Unicode codepoints like `"\u1234"` in the display, that the bytes show up encoded as bytes like "`\0xE1\0x88\0xB4`".  The actual underlying data is no different - any UTF-8 character can be treated as 1 to 4 bytes.

kevpl:
To follow up on @Iristyle's Beaker mention, Beaker currently supports Ruby 1.9, which is only supported in the [2.9.x net-ssh series](http://net-ssh.github.io/net-ssh/#label-RUBY+SUPPORT). We are, however, looking to leave Ruby 1.9 support behind soon, and will look at what we can do to get onto later versions then.

mfazekas:
@Iristyle i'm thinking about something along the lines of https://github.com/net-ssh/net-ssh/pull/402

mfazekas:
Thanks for the patch again, this should be now fixed in 4.0.0.beta1

mfazekas:
BTW there is String#b in ruby 2.0+
http://ruby-doc.org/core-2.0.0/String.html#method-i-b
that should be the one to use instead of ```dup.force_encoding("ASCII-8BIT")```

PR #407, opened 2016-07-23
Utf8

mfazekas:
Closes: #394


Iristyle:
Any chance this could be backported to the 3.x series?

PR #431, opened 2016-09-02
gemspec fixes for jruby sepcifics

jwoertink:
This gem doesn't install with JRuby because of the required Ruby version. It's looking for MRI > 2.0. This patch removes that dependency for MRI, and also fixes the `jruby-pageant` dependency. `RUBY_PLATFORM` returns 'java' for JRuby, so I changed it to `RUBY_ENGINE`.

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -105,7 +105,7 @@ If you don't add the public key, you'll see an error like "Couldn't verify data 
 
 * Ruby 1.8.x is supported up until the net-ssh 2.5.1 release.
 * Ruby 1.9.x is supported up until the net-ssh 2.9.x release.
-* Current net-ssh releases require Ruby 2.0 or later.
+* Current net-ssh releases require Ruby 2.0 or later. JRuby 1.7+ (with `compat.version=2.0` set) or JRuby 9k+
 
 == RUNNING TESTS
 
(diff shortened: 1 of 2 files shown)


mfazekas:
Are you sure the ```required_ruby_version``` doesn't work on jruby? Which jruby exactly? What is the exact error? IMHO it should work, and we want that on jruby too as we need a jruby which supports ruby 2.0+ features.

jwoertink:
When I try to install it fails with
```
[23:39PM] docum3nt (master)$ gem install net-ssh
ERROR:  Error installing net-ssh:
	net-ssh requires Ruby version >= 2.0.
```

I tried looking for what the jruby equivalent was, but the only thing I saw was setting the platform to "java". The only thing with that though is when you set the platform to java, that means the gem is a jruby only gem which we wouldn't want. 

I'm actually not using the latest jruby because of other issues I was running in to. Let me try a few versions, and see what I can find out.

thomas0087:
I was having the same issue, I upgraded my project to use jruby 9k and the gem installs without needing this PR.

jwoertink:
Ok, so it turns out that specifying the version works fine with JRuby 1.7+ if you have `compat.version=2.0` set. This pull request now just documents this in the README, as well as fixes the `RUBY_ENGINE` for the jruby dependency. The alternative is to use `RUBY_PLATFORM == "java"` for that dependency.

PR #432, opened 2016-09-09
Support HTTP Proxy on SSL socket

milgner:
Since the HTTP Proxy only supports BASIC authentication, this adds the required infrastructure to connect to a SSL-enabled server. I kept the changes as local as possible and ran `rubocop` to check for style consistency.

I could not locate any tests for the HTTP proxy so I wasn't sure where to add one for HTTPS :confused:


lib/net/ssh/proxy/http.rb
```diff
@@ -68,7 +67,12 @@ def open(host, port, connection_options)
       raise ConnectError, resp.inspect
     end
 
-    private
+    protected
```
mfazekas: i think ```private``` should work

mfazekas:
Added test for httpproxy - see #435

PR #436, opened 2016-10-11
detach proxycommand processes to avoid them becoming zombies

willbryant:
In Unix/Linux/etc, you must call one of the wait methods (eg. Process.wait) to retrieve the exit code of a terminated process.  Until you do this, the process sits in the process table as a zombie process, and over time as a number of connections get made and closed through a proxycommand-style proxy they pile up like this:

$ pstree|grep -F '(ssh)'|wc -l
      12
$ pstree|grep -F '(ssh)'|wc -l
      15
$ pstree|grep -F '(ssh)'|wc -l
      28

Since we have no interest in the exit code of the proxycommand process (we have never checked it before), we can detach the process instead, so that it doesn't becomes a zombie (technically the PID 1 process waits for it instead).

We still have open IO handles to the process, so there's no change to how long that process lives.

### DIFF
--- a/lib/net/ssh/proxy/command.rb
+++ b/lib/net/ssh/proxy/command.rb
@@ -56,6 +56,7 @@ module Net; module SSH; module Proxy
       }
       begin
         io = IO.popen(command_line, "r+")
+        Process.detach(io.pid)
         if result = Net::SSH::Compat.io_select([io], nil, [io], 60)
           if result.last.any?
             raise "command failed"


mfazekas:
See #378

The io returned by the proxy command should be closed with ```.close```, so they should not became zombies. Do you have a reproducible way to make zombies with proxy command?

mfazekas:
will be reopening once we know whats going on

willbryant:
Yes, it's easy to reproduce - just make 100 connections to a host using a proxy command for each one, wait for all the processes to exit *without exiting the parent process*, and look in `ps`.  You will see the zombie processes marked in brackets like "(ssh)".

I think you are confusing whether the process exits with whether the process stays in the process table.

This is a unix thing.  You must either detach each child process, or wait for its exit code - whether or not you are interested in the exit code.

willbryant:
Documented here, for example: https://en.wikipedia.org/wiki/Zombie_process

mfazekas:
@willbryant If you use net-ssh correctly (closing net-ssh session) it should close the IO object returned by IO.popen. That will not leave any zombie process behind.

This
```ruby
p = IO.popen('echo hello')
puts p.read()
p.close()
```
will not leave any zombie process. Nor should correct use of net-ssh. 

If you disagree, send concrete testcase (ruby code). That's something we can discuss, i don't have time for anything else.

PR #456, opened 2016-11-28
#357 as a special case, ignore nil :password options, to avoid breaki…

mfazekas:
…ng capistrano v2.

Closes: #455


mfazekas:
@willbryant thanks much!
This is now available in 4.0.0.rc1
https://rubygems.org/gems/net-ssh/versions/4.0.0.rc1

Issue #458, opened 2016-12-07
Passphrase for ssh key can't be nil (bug)


detatka-kuzlatka-otevrete:
Please apply the attached fix for ssh keys without a passphrase, now the check in Net::SSH.start raises:
ArgumentError, "Value(s) have been set to nil: passphrase"

[net-ssh.diff.txt](https://github.com/net-ssh/net-ssh/files/636443/net-ssh.diff.txt)

mfazekas:
@detatka-kuzlatka-otevrete what is you net-ssh version? what is the usecase? why are you setting the passsphrase to nil?

detatka-kuzlatka-otevrete:
my net-ssh version is: 4.0.0.rc1

the usecase is: i am using netconf gem:
https://github.com/Juniper/net-netconf/blob/master/lib/net/netconf/ssh.rb#L28

i am not setting the passphrase to nil.

mfazekas:
Thanks for the report

This was a change in 3.0 => 4.0 it can be fixed on net-netconf side too. AFAIK there is no reason to set the passphrase to nil

Sod-Almighty:
You are mistaken. There are plenty of reasons to set things to nil.

Here's an excerpt from my code, that no longer works thanks to the _entirely unnecessary_ guard clause in 4.0:

```
Net::SSH.start @address, @user, key_data: @key && [@key],
  password: !@key && @password, passphrase: @key && @password,
  config: false, non_interactive: true
```

Result:
> Value(s) have been set to nil: key_data, passphrase (ArgumentError)

What you're essentially forcing me to do is assemble an options hash explicitly, then pass it to the call:
```
opts = { config: false, non_interactive: true }
opts[:key_data] = [@key] if @key
opts[:password] = @password unless @key
opts[:passphrase] = @password if @key
Net::SSH.start @address, @user, **opts
```
**How is this helpful?**

For that matter, what purpose is served by the guard clause in the first place?

mfazekas:
@Sod-Almighty I'd appreciate if you'd choose your words more carefully. This is an open source project where people are spending they valuable free time. It's possible and likely that are wrong decisions made, but please focus on understanding the decisions then explaining why they are wrong.

See #355, #316


If you still want to do in one line you can do this:
```ruby
Net::SSH.start @address, @user, {key_data: @key && [@key],
  password: !@key && @password, passphrase: @key && @password,
  config: false, non_interactive: true}.reject { |_,v| v.nil? }
```

[...]

PR #465, opened 2016-12-15
Allow nil for passphrase, warn instead of raise with nil option errors

mfazekas:
Fixes: #458


mfazekas:
@KaneMorgan pls review

Issue #314, opened 2016-02-28
correctly parse "+" in config files 


RyanJarv:
Running the following with "HostkeyAlgorithms +ssh-dss" set in ~/.ssh/config will result in a "Net::SSH::Exception: could not settle on host_key algorithm" error when connecting to an ubuntu14.04 instance (haven't tested against any other nodes). 

Net::SSH.start('1.2.3.4', 'ubuntu', {:timeout=>10, :auth_methods=>["publickey"], :keys_only=>true, :host_key_alias=>"host_key_alias", :key_data=> ["ssh_private_key"]})

This is only an issue with net-ssh version 3.0.2, when using 2.9.2 or using OpenSSH I am able to connect without issue. Removing the "HostkeyAlgorithms +ssh-dss" from my ssh config also fixes the issue as long as you are using a rsa key, dss is deprecated on newer versions of OpenSSH which makes this line necessary in some cases.

mfazekas:
Can you please add a :verbose=>:debug and post log?

RyanJarv:
Yep, here you go.

```
irb(main):006:0> Net::SSH.start('host.name', 'user', {:timeout=>10, :auth_methods=>["publickey"], :keys_only=>true, :host_key_alias=>"host_key_alias", :key_data=> ["ssh_private_key"], :verbose => :debug})
D, [2016-02-28T16:44:27.680672 #64621] DEBUG -- net.ssh.transport.session[3fc4ca1480ec]: establishing connection to hydra01a-dev:22
D, [2016-02-28T16:44:27.917419 #64621] DEBUG -- net.ssh.transport.session[3fc4ca1480ec]: connection established
I, [2016-02-28T16:44:27.917546 #64621]  INFO -- net.ssh.transport.server_version[3fc4ca145694]: negotiating protocol version
D, [2016-02-28T16:44:28.010825 #64621] DEBUG -- net.ssh.transport.server_version[3fc4ca145694]: remote is `SSH-2.0-OpenSSH_6.6.1p1 Ubuntu-2ubuntu2.3'
D, [2016-02-28T16:44:28.010926 #64621] DEBUG -- net.ssh.transport.server_version[3fc4ca145694]: local is `SSH-2.0-Ruby/Net::SSH_3.0.2 x86_64-darwin12.0'
[...]

PR #470, opened 2016-12-17
Correctly parse '+' in config files

clupprich:
Because I recently was also hit by https://github.com/net-ssh/net-ssh/issues/314, I wanted to help to get this feature into `Net::SSH`.

Two things I'm not totally sure about:

- when running `ssh -Q key` on my machine, I got a couple of algorithms that are not in the list of default ones. I guess that's fine, and that's also why I skipped validation for the algorithms added via `+` to the list
- I failed to find a reference (or lookup the source code of other implementations), but I think from the wording the `HostkeyAlgorithms` should only feature a single `+` sign, so that the whole line is added (vs a plus sign in front of each algorithm)

Let me know what you think - happy to update the PR!

Closes: #470

### DIFF
--- a/lib/net/ssh/transport/algorithms.rb
+++ b/lib/net/ssh/transport/algorithms.rb
@@ -212,35 +212,8 @@ module Net; module SSH; module Transport
       def prepare_preferred_algorithms!
         options[:compression] = %w(zlib@openssh.com zlib) if options[:compression] == true
 
-        ALGORITHMS.each do |algorithm, list|
-          # apply the preferred algorithm order, if any
-          if options[algorithm]
-            option = Array(options[algorithm]).compact.uniq
-
-            if option.first && option.first.start_with?('+')
-              algorithms[algorithm] = list.dup
-              algorithms[algorithm] << option.first[1..-1]
-              algorithms[algorithm].concat(option[1..-1])
-              algorithms[algorithm].uniq!
-            else
-              algorithms[algorithm] = option
-
-              if options[:append_all_supported_algorithms]
-                list.each { |name| algorithms[algorithm] << name unless algorithms[algorithm].include?(name) }
-              end
-            end
-
-            unsupported = []
-            algorithms[algorithm].select! do |name|
-              supported = ALGORITHMS[algorithm].include?(name)
-              unsupported << name unless supported
-              supported
-            end
-
-            lwarn { "unsupported #{algorithm} algorithm: `#{unsupported}'" } unless unsupported.empty?
-          else
-            algorithms[algorithm] = list.dup
-          end
+        ALGORITHMS.each do |algorithm, supported|
+          algorithms[algorithm] = compose_algorithm_list(supported, options[algorithm], options[:append_all_supported_algorithms])
         end
 
         # for convention, make sure our list has the same keys as the server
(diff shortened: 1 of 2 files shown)


lib/net/ssh/transport/algorithms.rb
```diff
@@ -213,22 +213,32 @@ def prepare_preferred_algorithms!
...
-              unsupported << name unless supported
-              supported
-            end
-            lwarn { "unsupported #{algorithm} algorithm: `#{unsupported}'" } unless unsupported.empty?
-
-            if options[:append_all_supported_algorithms]
-              list.each { |name| algorithms[algorithm] << name unless algorithms[algorithm].include?(name) }
+            optionals = Array(options[algorithm]).compact.uniq
+
+            if optionals.first.start_with?('+')
```
mfazekas: we should be still filtering the unsupported algs. We don't want to advertise something we don't support.

clupprich:
I just commited the change as requested. However, I'm thinking of re-doing this change with your solution in mind, as I think there's a difference between the default algorithms, and the available ones (which is also briefly mentioned in the `ssh_config` link you included). There could be an algo which is available, but it's not default (like the `ecdsa-sha2-nistp256-cert-v01@openssh.com` one I had in my tests).

mfazekas:
@clupprich looks better, not that we don't really have notion of default  algorithms, we only have list of supported ones.

About```ecdsa-sha2-nistp256-cert-v01@openssh.com``` i don't think is something we support at the moment.

Also according to ssh docs order is important. 
>  Specifies the key types that will be used for hostbased authentication as a comma-separated pattern list.  Alternately if the specified value begins with a `+' character, then the specified key types will be appended to the default set instead of replacing them. 

So ideally we should have the same logic of prepending default alg before them. 
Which means the '+' is kind of pointless at the moment as if our default is every supported then what's after '+' is already included or unsupported.

clupprich:
> Which means the '+' is kind of pointless at the moment as if our default is every supported then what's after '+' is already included or unsupported.

Yep, I agree. But at least we're no longer stuck with no algorithms ;)

I'm happy to draft another PR that will bring in default vs supported algorithms.

mfazekas:
@clupprich  yes i agree having non empty is way better. Not sure if it's a good idea to put default args to 4.0, otherwise default args would be nice.

mfazekas:
So if we want default let's add it asap, so we can have it with 4.0 otherwise it should go with 5.0, we don't want to introduce defaults in 4.1 or so.
If we don't want default the this should work great.

lib/net/ssh/transport/algorithms.rb
```diff
@@ -213,22 +213,33 @@ def prepare_preferred_algorithms!
...
 
         ALGORITHMS.each do |algorithm, list|
-          algorithms[algorithm] = list.dup
-
           # apply the preferred algorithm order, if any
           if options[algorithm]
-            algorithms[algorithm] = Array(options[algorithm]).compact.uniq
+            option = Array(options[algorithm]).compact.uniq
+
+            if option.first.start_with?('+')
```
mfazekas: we either need to handle that option with zero element here or make sure it's not possible on all the code paths calling this.

clupprich: I think we're good here: If the option is `nil`, we're not entering that section at all. If it's something else, `Array()` will always make sure we're dealing with an array.

mfazekas: Yeah but if it's an empty array ```option.first``` will be nil which doesn't responds to ```starts_with?```

clupprich: With `Array()`, that would only happen with `nil`, as far as I know.

mfazekas: I mean ```options[algorithm]==[]```

```irb
2.3.1 :001 > option = Array([]).compact.uniq
 => [] 
2.3.1 :002 > option.first.start_with?('+')
NoMethodError: undefined method `start_with?' for nil:NilClass
	from (irb):2
	from /Users/boga/.rvm/rubies/ruby-2.3.1/bin/irb:11:in `<main>'
```

clupprich: You're right!

mfazekas:
Looks great! 
Only thing we need is fixing rubocop issue with complexity. We should probably extract the body of  ```ALGORITHM.eac .. ``` part to a private method.

like this
```ruby
ALGORITHMS.each do |algorithm, list|
  algorithms[algorithm] = preferred_algorithm_list(options[algorithm], list, append_all_supported_algorithms : options[:append_all_supported_algorithms])
end
```

but i'm sure there is a better name than preferred_algorithm_list


```
9.59s$ bundle _1.11.2_ exec rubocop
Inspecting 153 files
..................................................C......................................................................................................
Offenses:
lib/net/ssh/transport/algorithms.rb:211:7: C: Assignment Branch Condition size for prepare_preferred_algorithms! is too high. [81.64/76]
      def prepare_preferred_algorithms!
      ^^^
153 files inspected, 1 offense detected
The command "bundle _1.11.2_ exec rubocop" exited with 1.
```

clupprich:
Weird that I didn't see the violation in my editor (Atom) - are you running modified rubocop config files in CI?

mfazekas:
@clupprich thanks much! 👍

clupprich:
Sure thing! 😀

PR #475, opened 2016-12-28
Support for `Include` directive

masa0x80:
Net::SSH supports for `Include` keyword since OpenSSH added an include directive for ssh_config on v7.3.
Ref: https://www.openssh.com/txt/release-7.3

Include directives are very useful for breaking down a large ssh_config into smaller files.

```
$ man ssh_config
...
     Include
             Include the specified configuration file(s).  Multiple pathnames may be
             specified and each pathname may contain glob(3) wildcards and, for user
             configurations, shell-like ``~'' references to user home directories.
             Files without absolute paths are assumed to be in ~/.ssh if included in a
             user configuration file or /etc/ssh if included from the system configu-
             ration file.  Include directive may appear inside a Match or Host block
             to perform conditional inclusion.
...
```

### DIFF
--- a/lib/net/ssh/config.rb
+++ b/lib/net/ssh/config.rb
@@ -73,6 +73,7 @@ module Net; module SSH
       # options.)
       def load(path, host, settings={})
         file = File.expand_path(path)
+        base_dir = File.dirname(file)
         return settings unless File.readable?(file)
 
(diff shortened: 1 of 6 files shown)


mfazekas:
👍 
thanks much for the contribution

PR #481, opened 2017-01-03
Always use Net::SSH::Config.expandable_default_files

mcasper:
#351 partially fixed the expandable default files problem, but
Net::SSH.configuration_for still passes in the unfiltered default files,
bypassing the filtered defaults. This change exposes the
`expandable_default_files` method to allow Net::SSH to fall back to it,
to avoid the error:
```
ArgumentError: couldn't find HOME environment -- expanding `~'
```


mfazekas:
👍  Thanks for the PR!

PR #485, opened 2017-01-10
Add support for SSH certificates for client authentication.

dtbartle:
This adds support for parsing and signing SSH certificates, and for using
SSH certificates as identities when authenticating, via a "-cert.pub" file.

This does not add support for authenticating host certificates.  Supporting
that requires understanding @cert-authority in known_hosts, and validating
the hostname against the certificate principals.

### DIFF
--- a/lib/net/ssh/authentication/certificate.rb
+++ b/lib/net/ssh/authentication/certificate.rb
@@ -1,0 +1,170 @@
+require 'net/ssh/buffer'
+require 'securerandom'
+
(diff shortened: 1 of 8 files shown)


dtbartle:
Fixed to work with ruby 2.0.

mfazekas:
👏  thanks much for the contribution. Looks great!

mfazekas:
This is awesome it works great! Thanks again. I've added an integration test for this at #486

PR #487, opened 2017-01-12
Add support for adding and removing identities from SSH agents.

dtbartle:


lib/net/ssh/authentication/agent.rb
```diff
@@ -139,6 +147,37 @@ def sign(key, data)
       return reply.read_string
     end
 
+    # Adds the private key with comment to the agent.
+    # If opts[:lifetime] is given, the key will automatically be removed after
+    # opts[:lifetime] seconds.
+    # If opts[:confirm] is true, confirmation will be required for each agent
+    # signing operation.
+    def add_identity(priv_key, comment, opts={})
```
mfazekas: instead of opt, we could use kw arguments? what do you think? We now require ruby 2.0.

dtbartle: Sounds good; updated the PR.

mfazekas:
Thanks for the quality PR again!

dtbartle:
@mfazekas I noticed https://github.com/net-ssh/net-ssh/pull/487/commits/09abd0afea40fda90262311825167754fa9e0ecf didn't get merged; should I send a separate PR for that?

PR #488, opened 2017-01-17
Fix libsodium errors

chapmajs:
This PR fixes two issues that happen with the system does not provide `libsodium` and you're using ed25519 keys:

* Fixes whiny nil when the error is not a `Gem::LoadError` (string left uninitialized, next line pushes onto nil)
* Adds better error message when `rbnacl` is installed but the system doesn't provide `libsodium`

I can smash it down to a single commit but wanted to leave the two separate for clarity.

### DIFF
--- a/lib/net/ssh/authentication/ed25519.rb
+++ b/lib/net/ssh/authentication/ed25519.rb
@@ -1,7 +1,12 @@
 gem 'rbnacl', '>= 3.2.0', '< 5.0'
 gem 'bcrypt_pbkdf', '~> 1.0' unless RUBY_PLATFORM == "java"
 
-require 'rbnacl/libsodium'
+begin
+  require 'rbnacl/libsodium'
+rescue LoadError => e
+  raise LoadError.new(e.message + ', your system does not provide libsodium, install the rbnacl-libsodium gem')
+end
+
 require 'rbnacl'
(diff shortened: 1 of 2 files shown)


mfazekas:
Thanks much for the contribution. On ```require 'rbnacl/libsodium'``` i'm not 100% we do need that line. `rbnacl/libsodium` is an optional dependency and some systems - linux? can have sodium preinstalled and we should be able to work without rbnacl-libsodium gem at all.

Like this description - https://github.com/cryptosphere/rbnacl/wiki/Windows-Installation doesn't require libsodium

chapmajs:
I believe when I removed the require `'rbnacl/libsodium'` initially (supposed to only need it for passphrase protected ed25519 keys) that `require rbnacl` blew up with more unhelpful errors.

mfazekas:
@chapmajs many thanks for the PR, i've merged a modified version where i've removed `require 'rbnacl/libsodium',`  and catch `LoadError` instead of `Gem::LoadError` to explain dependencies needed

chapmajs:
Thanks for taking care of this! Happy to help!

mfazekas:
@chapmajs 4.1.0.beta1 was released ideally it should work without rbnacl/libsodium but should give you reasonable error message if it's not installed and no libsodium found.

PR #490, opened 2017-01-18
Have Authentication Sessions not use defaults when key_data option is present

joshlarson:
An attempt to fix https://github.com/net-ssh/net-ssh/issues/489 by only adding `default_keys` to the `key_manager` if both `keys` and `key_data` are empty.

### DIFF
--- a/lib/net/ssh/authentication/key_manager.rb
+++ b/lib/net/ssh/authentication/key_manager.rb
@@ -182,6 +182,10 @@ module Net
           nil
         end
 
+        def no_keys?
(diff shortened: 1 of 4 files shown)


mfazekas:
👍 thanks much for the PR, i like it if key_data is provided we should not try default keys (or only try after)
There is error in 2.4 tests and jruby tests, have looked into them?

joshlarson:
@mfazekas I haven't looked into them yet, but I'll try to get them passing. Thanks for the positive feedback! 😄

PR #493, opened 2017-01-20
Don't try to keep sending buffered data on a closed socket.

coderanger:
This happens to me with net-sftp during `close_channel`, not sure exactly why. Tracing in `enqueue_message` the socket is open at the time the packet is buffered and then it gets closed part way through a processing loop. This might be the wrong fix, but I'm not super sure what else to fix specifically.

### DIFF
--- a/lib/net/ssh/connection/session.rb
+++ b/lib/net/ssh/connection/session.rb
@@ -268,7 +268,7 @@ module Net; module SSH; module Connection
       end
 
       Array(writers).each do |writer|
-        writer.send_pending
+        writer.send_pending unless writer.closed?
       end
     end
 


mfazekas:
@coderanger thanks much for the PR!

Would it be possible to came up with a way to reproduce the issue (ideally without `chef`)? That would be really usefull. I'm also not sure if this is the correct fix, but to decide that i need to understand the issue first!

coderanger:
Unfortunately I don't think I'll be able to, it was already hard to narrow down the repro even within Test Kitchen :-( I'll give a shot at a more minimal repro case tomorrow though, at least I can narrow down some of the trigger issues since it wasn't even happening on all my cookbooks.

mfazekas:
If you can send repro with test kitchen with description on how to set up that might work as well.

coderanger:
So the consistent repro I was using for testing was in https://github.com/poise/application_javascript, add this to the Gemfile and then `bundle up`:

```ruby
gem 'chef', github: 'chef/chef'
gem 'kitchen-sync', '2.1.1' # I pushed a 2.1.2 of this gem with a rescue IOError to mask the issue
```

Then this should trigger it `bundle exec env POISE_MASTER_BUILD=true kitchen test ubuntu`. It will take about 5 minutes on my machine, though for you it might be a bit longer because it would default to using Vagrant for virtualization. I can email you directly a key to let you use my build system so you don't have to deal with setting up Vagrant and all that lovely stuff. This is still a lot to do so it's cool if you want to wait until morning when I can try and rig up a standalone case.

mfazekas:
@coderanger i could not repro the issue.

Gemfile is like this:

```ruby
    gem name, github: github
  end
end

gem 'chef', github: 'chef/chef'
gem 'kitchen-sync', '2.1.1' 

dev_gem 'halite'
dev_gem 'poise'
...
```

mfazekas:

coderanger:
I mean is there any reason to not merge this?

mfazekas:
@coderanger Sorry i cannot merge this as i'd need more information on this issue. Ideally a reproduction but at least a stacktrace.

PR #495, opened 2017-01-25
Fix agent detection on Windows

ChristianKoehler:
The agent detection fails if you have a mingw setup with a working ssh-agent. The ENV SSH_AUTH_SOCK is set, but windows fails to use the UnixSocket class.

```
E, [2017-01-25T09:46:15.783805 #75060] ERROR -- net.ssh.authentication.agent[24dcda0]: could not connect to ssh-agent: uninitialized constant Net::SSH::Authentication::Agent::UNIXSocket
Did you mean?  Socket
               IPSocket
               UDPSocket
```

Tested on:
```
$ ruby --version
ruby 2.3.3p222 (2016-11-21 revision 56859) [i386-mingw32]
```

### DIFF
--- a/lib/net/ssh/authentication/agent.rb
+++ b/lib/net/ssh/authentication/agent.rb
@@ -71,7 +71,7 @@ module Net; module SSH; module Authentication
       @socket =
         if agent_socket_factory
           agent_socket_factory.call
-        elsif ENV['SSH_AUTH_SOCK'] && defined?(unix_socket_class)
+        elsif ENV['SSH_AUTH_SOCK'] && unix_socket_class
           unix_socket_class.open(ENV['SSH_AUTH_SOCK'])
         elsif Gem.win_platform? && RUBY_ENGINE != "jruby"
           Pageant::Socket.open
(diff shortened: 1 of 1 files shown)


lib/net/ssh/authentication/agent.rb
```diff
@@ -71,10 +71,10 @@ def connect!(agent_socket_factory = nil)
       @socket =
         if agent_socket_factory
           agent_socket_factory.call
-        elsif ENV['SSH_AUTH_SOCK'] && defined?(unix_socket_class)
```
mfazekas: I don't think we should reverse these lines, the only fix we need is either removing `unix_socket_class` and using `UNIXSocket` instead or changing 

```ruby
def unix_socket_class
     UNIXSocket
end
```

to 

```ruby
 def unix_socket_class
       defined?(UNIXSocket) && UNIXSocket
end
```

and using `unix_socket_class` instead of `defined?(unix_socket_class)`

ChristianKoehler: Your proposed fix works like a charm. Should I update the PR, or do you want to fix it by your own? My ruby skills are not good enough to come up with a test case for this.

PR #497, opened 2017-02-08
Add ProxyJump support

### LINKED ISSUE #430: Could you please add support for the ProxyJump configuration?
ProxyJump makes it much simpler to set up a jump host. Here are some references:

http://www.openssh.com/txt/release-7.3

```
New Features
------------

 * ssh(1): Add a ProxyJump option and corresponding -J command-line
   flag to allow simplified indirection through a one or more SSH
   bastions or "jump hosts".
```

https://en.wikibooks.org/wiki/OpenSSH/Cookbook/Proxies_and_Jump_Hosts#Passing_Through_One_or_More_Gateways_Using_ProxyJump

https://wiki.gentoo.org/wiki/SSH_jump_host

```
#~/.ssh/configProxyJump Example

### First jumphost. Directly reachable
Host betajump
  HostName jumphost1.example.org

### Host to jump to via jumphost1.example.org
Host behindbeta
  HostName behindbeta.example.org
  ProxyJump  betajump
```

rmm5t:
Closes #430 

This implementation was heavily based on the OpenSSH implementation of `ProxyJump`:
https://github.com/openssh/openssh-portable/blob/dda78a03af32e7994f132d923c2046e98b7c56c8/ssh.c#L1096

It's worth noting that this implementation ironically depends on an underlying OpenSSH command-line installation. But if you already had a reliance on `ProxyCommand` or `ProxyJump` ssh_config directives, that dependency should already be met. I suppose, ideally, this `ProxyJump` implementation might later be better implemented as a native Net::SSH proxy without having to rely on an `ssh` executable, but the recursive complexities of that might outweigh the benefits.

### DIFF
--- a/lib/net/ssh/config.rb
+++ b/lib/net/ssh/config.rb
@@ -27,6 +27,7 @@ module Net; module SSH
   # * Port => :port
   # * PreferredAuthentications => maps to the :auth_methods option
   # * ProxyCommand => maps to the :proxy option
+  # * ProxyJump => maps to the :proxy option
   # * PubKeyAuthentication => maps to the :auth_methods option
   # * RekeyLimit => :rekey_limit
   # * User => :user
@@ -224,6 +225,11 @@ module Net; module SSH
             if value and !(value =~ /^none$/)
               require 'net/ssh/proxy/command'
               hash[:proxy] = Net::SSH::Proxy::Command.new(value)
+            end
+          when 'proxyjump'
+            if value
+              require 'net/ssh/proxy/jump'
+              hash[:proxy] = Net::SSH::Proxy::Jump.new(value)
             end
           when 'pubkeyauthentication'
             if value
--- a/lib/net/ssh/proxy/jump.rb
+++ b/lib/net/ssh/proxy/jump.rb
@@ -1,0 +1,53 @@
+require 'uri'
+require 'net/ssh/proxy/command'
+
+module Net; module SSH; module Proxy
+
+  # An implementation of a jump proxy. To use it, instantiate it,
+  # then pass the instantiated object via the :proxy key to
+  # Net::SSH.start:
+  #
+  #   require 'net/ssh/proxy/jump'
+  #
+  #   proxy = Net::SSH::Proxy::Jump.new('user@proxy')
+  #   Net::SSH.start('host', 'user', :proxy => proxy) do |ssh|
+  #     ...
+  #   end
+  class Jump < Command
+
+    # The jump proxies
+    attr_reader :jump_proxies
+
+    # Create a new socket factory that tunnels via multiple jump proxes as
+    # [user@]host[:port].
+    def initialize(jump_proxies)
+      @jump_proxies = jump_proxies
+    end
+
+    # Return a new socket connected to the given host and port via the jump
+    # proxy that was requested when the socket factory was instantiated.
+    def open(host, port, connection_options = nil)
+      build_proxy_command_equivalent(connection_options)
+      super
+    end
+
+    # We cannot build the ProxyCommand template until we know if the :config
+    # option was specified during `Net::SSH.start`.
+    def build_proxy_command_equivalent(connection_options = nil)
+      first_jump, extra_jumps = jump_proxies.split(",", 2)
+      config = connection_options && connection_options[:config]
+      uri = URI.parse("ssh://#{first_jump}")
+
+      template = "ssh"
+      template << " -l #{uri.user}"    if uri.user
+      template << " -p #{uri.port}"    if uri.port
+      template << " -J #{extra_jumps}" if extra_jumps
+      template << " -F #{config}" if config != true && config
+      template << " -W %h:%p "
+      template << uri.host
+
+      @command_line_template = template
+    end
+  end
+
+end; end; end
--- a/test/configs/proxy_jump
+++ b/test/configs/proxy_jump
@@ -1,0 +1,5 @@
+Host behind-proxy
+  ProxyJump user@proxy
+
+Host behind-three-proxies
+  ProxyJump user1@proxy1,user2@proxy2,user3@proxy3
--- a/test/integration/test_proxy.rb
+++ b/test/integration/test_proxy.rb
@@ -4,6 +4,7 @@ require 'net/ssh'
 require 'timeout'
 require 'tempfile'
 require 'net/ssh/proxy/command'
+require 'net/ssh/proxy/jump'
 
 class TestProxy < NetSSHTest
   include IntegrationTestHelpers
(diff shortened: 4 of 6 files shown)


mfazekas:
@rmm5t looks awesome! Thanks much! 

Ideally we'd also need an integartion test - https://github.com/net-ssh/net-ssh/tree/master/test/integration

rmm5t:
> Ideally we'd also need an integration test...

Ah! Thanks for the pointer to the integration tests README. I wondered how those tests were managed. I can look into adding that maybe later this week; however, the implementation is just a layer on top of `Net::SSH::Proxy::Command`, which already has integration tests, and there's a unit test to verify that `Proxy::Jump` is just a `Proxy::Command`.

Aside: this implementation was heavily based on the OpenSSH implementation of `ProxyJump`:
https://github.com/openssh/openssh-portable/blob/dda78a03af32e7994f132d923c2046e98b7c56c8/ssh.c#L1096

mfazekas:
If you can describe a quick setup with localhost i should be able to add a test for it. We have a few users with known password and keys, and i assume we can add entries to /etc/hosts if we need aliases for hostnames.

rmm5t:
@mfazekas An integration test would look very similar to the integration test for `Proxy::Command`:
https://github.com/net-ssh/net-ssh/blob/master/test/integration/test_proxy.rb#L33-L42

I just added an integration test to `test/integration/test_proxy.rb`. I don't have the time to setup an environment to ensure that integration test runs locally, but it should be a good start (...hoping appveyor tests just pass on this first attempt...).

rmm5t:
@mfazekas Is there something I need to do to make the Travis CI's integration tests pass, or is a failure expected when it runs from my fork's branch?

mfazekas:
`bundle exec rubocop` does errors that you should be able to run locally too

then there is some issue with integration tests i'll look into it tomorrow if it's not obvious to you what's the issue

```
  1) Error:
TestProxy#test_proxy_jump_through_localhost:
Errno::EPIPE: Broken pipe
    /home/travis/build/net-ssh/net-ssh/lib/net/ssh/transport/server_version.rb:44:in `write'
    /home/travis/build/net-ssh/net-ssh/lib/net/ssh/transport/server_version.rb:44:in `negotiate!'
    /home/travis/build/net-ssh/net-ssh/lib/net/ssh/transport/server_version.rb:32:in `initialize'
    /home/travis/build/net-ssh/net-ssh/lib/net/ssh/transport/session.rb:84:in `new'
    /home/travis/build/net-ssh/net-ssh/lib/net/ssh/transport/session.rb:84:in `initialize'
    /home/travis/build/net-ssh/net-ssh/lib/net/ssh.rb:233:in `new'
    /home/travis/build/net-ssh/net-ssh/lib/net/ssh.rb:233:in `start'
    /home/travis/build/net-ssh/net-ssh/test/integration/test_proxy.rb:97:in `block in test_proxy_jump_through_localhost'
    /home/travis/build/net-ssh/net-ssh/test/integration/test_proxy.rb:30:in `block in setup_ssh_env'
    /home/travis/build/net-ssh/net-ssh/test/integration/common.rb:19:in `block in tmpdir'
    /home/travis/.rvm/rubies/ruby-2.0.0-p648/lib/ruby/2.0.0/tmpdir.rb:88:in `mktmpdir'
    /home/travis/build/net-ssh/net-ssh/test/integration/common.rb:18:in `tmpdir'
    /home/travis/build/net-ssh/net-ssh/test/integration/test_proxy.rb:25:in `setup_ssh_env'
    /home/travis/build/net-ssh/net-ssh/test/integration/test_proxy.rb:95:in `test_proxy_jump_through_localhost'
```

rmm5t:
> bundle exec rubocop does errors that you should be able to run locally too

The rubocop offenses are only happening because the existing implementation of the `translate` method that parses `~/.ssh/config` (huge `case` statement) already ran on the edge of rubocop's configured complexity threshold.  I only added a necessary `when` clause for `ProxyJump`, and I don't think a refactor to reduce this complexity metric is warranted or appropriate for this particular feature branch (i.e. it's unrelated and an arbitrary rubocop warning and  should be a separate  branch and PR).

> then there is some issue with integration tests i'll look into it tomorrow if it's not obvious to you what's the issue

Yeah, sorry, it's not immediately obvious to me, especially because the same integration tests appear to have passed on appveyor.

mfazekas:
Ok i'll take over and will add comment to silent rubocop and will debug integration test. Windows has different integration tests `win_integration` that's the reason why it's passsing.

mfazekas:
So the issue with integration test is that 

```ruby
def test_proxy_jump_through_localhost
  setup_ssh_env do
  proxy = Net::SSH::Proxy::Jump.new("#{user}@localhost")
    output = Net::SSH.start(*ssh_start_params(proxy: proxy)) do |ssh|
       ssh.exec! "echo \"$USER:echo123\""
     end
     assert_equal "net_ssh_1:echo123\n", output
   end
end
```

is that in this setup the key to allow to auth as `user@localhost` is `/tmp/.../id_rsa` aka `@key_id_rsa` see `setup_ssh_env` but that is not passed to the ssh command generated by proxy jump.

rmm5t:
> ...is that in this setup the key to allow to auth as user@localhost is /tmp/.../id_rsa aka @key_id_rsa see setup_ssh_env but that is not passed to the ssh command generated by proxy jump.

Ah! Good catch. Hmm. I wonder if this problem also exists in OpenSSH, because an overridden identity file is not passed through to the SSH command generated by their `ProxyJump` either.

mfazekas:
> Ah! Good catch. Hmm. I wonder if this problem also exists in OpenSSH, because an overridden identity file is not passed through to the SSH command generated by their ProxyJump either.

Are you sure it have to pass?
If i do 
`ssh -i key foo@bar.com -J bar@baz.com` 

does key used to auth `foo@bar.com` or does it used for both `foo@bar.com` and `bar@baz.com`?

rmm5t:
> Are you sure it have to pass?

Good point. I'm going to revert my last commit (it didn't work for the integration test scenario anyway).

> If i do `ssh -i key foo@bar.com -J bar@baz.com`, does key used to auth foo@bar.com or does it used for both foo@bar.com and bar@baz.com?

Good question. I'm not sure how to solve this at the moment. Might have to noodle on it.

mfazekas:
i've made those changes

- added `gateway.netssh` as alias to /etc/hosts
- i've added `gwkey` pubkey to user `net_ssh_2`
- i've added  this to ~/.ssh/config
   ```
    Host gateway.netssh
       IdentityFile gwkey.priv
   ```
- i've used `ssh -J net_ssh_2@gateway.netssh net_ssh_1@localhost -i mykey` equivalent

So the question is that in this config should we use gwkey.priv to auth to `net_ssh_2@gateway.netssh` or `mykey`?

see
https://github.com/net-ssh/net-ssh/pull/500/commits/37b051c6acfc2f66c7c1a852f603bf895e886408#diff-e4fa8581f0bf2840604e690cde65010eR34

mfazekas:
So this seems to work in way described with openssh too so dest host is authed with what's passed in via `-i` and jump host is authed with what's configured in `~/.ssh/config`

`ssh -i tmp_id_rsa net_ssh_1@localhost -J net_ssh_2@gateway.netssh -v`

mfazekas:
Merged in #500

rosenfeld:
Awesome! Thanks!

mfazekas:
FYI i've released 4.1.0rc1 https://rubygems.org/gems/net-ssh/versions/4.1.0.rc1

rmm5t:
@mfazekas Just tried it with capistrano, and it worked! Thanks!

PR #502, opened 2017-02-24
add direct-streamlocal@openssh.com support in Forward class

apachelogger:
- creates a unix domain socket on localhost
- requests a `direct-streamlocal@openssh.com` channel 
- sshd connects to remote unix domain socket

The channeling works exactly like `direct-tcpip`, only the "setup" part is different. This works with openssh 6.7+ https://github.com/openssh/openssh-portable/commit/7acefbbcbeab725420ea07397ae35992f505f702

I am a tad unsure how to properly test this. The current integration tests use ubuntu/trusty in vagrant, and trusty only had openssh 6.6. Options are probably to switch to ubuntu/xenial (16.04 LTS) or pretend there isn't much to test here (which actually is true given the only new code is creating a socket and the different channel type string).

Anyway, some input would be greatly appreciated.

### DIFF
--- a/lib/net/ssh/service/forward.rb
+++ b/lib/net/ssh/service/forward.rb
@@ -28,6 +28,7 @@ module Net; module SSH; module Service
       @remote_forwarded_ports = {}
       @local_forwarded_ports = {}
       @agent_forwarded = false
+      @local_forwarded_sockets = {}
 
       session.on_open_channel('forwarded-tcpip', &method(:forwarded_tcpip))
       session.on_open_channel('auth-agent', &method(:auth_agent_channel))
@@ -99,9 +100,7 @@ module Net; module SSH; module Service
       local_port
     end
 
-    # Terminates an active local forwarded port. If no such forwarded port
-    # exists, this will raise an exception. Otherwise, the forwarded connection
-    # is terminated.
+    # Terminates an active local forwarded port.
     #
     #   ssh.forward.cancel_local(1234)
     #   ssh.forward.cancel_local(1234, "0.0.0.0")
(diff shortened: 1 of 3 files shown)


lib/net/ssh/service/forward.rb
```diff
@@ -28,6 +28,7 @@ def initialize(session)
       @remote_forwarded_ports = {}
       @local_forwarded_ports = {}
       @agent_forwarded = false
+      @local_forwarded_sockets = {}
```
mfazekas: we don't seems to use this, do we?

apachelogger:
Yep, installing an own sshd is the way to go for testing then.

Should I be concenred that the the playbook.yml doesn't actually work with ansible 2.2.1.0 and rvm_io.ruby 2.0.0? First it seems to fall over permissions because sudo elevation isn't working and then installing gems for 2.0.0 fails because some of the gems that the playbook wants to install need a later ruby version.

lib/net/ssh/service/forward.rb
apachelogger: Now properly used. I've also added a test and adjusted the playbook to provision our own sshd, whether that works as expected on travis remains to be seen ^^

mfazekas:
#504 should fix the issue with ansible provisioning on vagrant. 

It looks like for some reason on travis restarting openssh kills the connection, and travis cannot continue. We do restart sshd normally also, but sure what's the issue in this case.

apachelogger:
Thanks, the playbook works much better now.

I've changed the test set-up. Instead of replacing the system sshd the playbook is now installing the new sshd into /opt and the socket forward test runs the newer sshd on port 2200.

lib/net/ssh/service/forward.rb
```diff
@@ -120,6 +121,59 @@ def active_locals
...
+          session.stop_listening_to(channel[:socket])
+          channel[:socket].close
+        end
+      end
+
+      local_socket_path
+    end
+
+    # Terminates an active local forwarded socket. If no such forwarded socket
+    # exists, this will raise an exception. Otherwise, the forwarded connection
```
mfazekas: i don't think it will raise an exception, will it? I'm fine both ways raising or not, but docs should state what the code is doing.

test/integration/test_forward.rb
```diff
@@ -644,4 +656,69 @@ def test_forward_local_unix_socket_to_remote_port
       end
     end
   end
+
+  def test_forward_local_unix_socket_to_remote_socket
+    setup_ssh_env do
+      start_sshd do |_pid, port|
+        session = Timeout.timeout(4) do
```
mfazekas: can we put this wait for port to live to `start_sshd`. We don't even have to connect with netssh just wait for the port to be alive until we can make a connection with `TCPSocket.new('localhost', port)`

test/integration/test_forward.rb
```diff
@@ -24,6 +24,18 @@
 class ForwardTestBase < NetSSHTest
   include IntegrationTestHelpers
 
+  # @yield [pid, port]
+  def start_sshd(port = '2200')
```
mfazekas: what about naming it somehting like `start_sshd_7_or_later` so it's more descriptive?

test/integration/test_forward.rb
apachelogger: That's what I had initially. Unfortunately, it doesn't work. I didn't look too much into why, but either sshd gets grumpy when you open a socket and close it again without any additional communication or _port-available_ doesn't actually mean sshd is ready to open a session.

At any rate, opening a session is the only reliable way I found.

apachelogger:
Changed the start_ssh method name and updated documentation to not claim raising exceptions when in fact nothing is being raised.

mfazekas:
@apachelogger thanks much for the contribution! great work! 👏

PR #509, opened 2017-03-21
New signature methods

whiteley:
These were added in https://github.com/openssh/openssh-portable/commit/76c9fbbe35aabc1db977fb78e827644345e9442e and available since OpenSSH_7.2p1.


mfazekas:
Nice, thanks much for the contribution. 

Can you show the usecase for this?

whiteley:
@mfazekas my use case is a bit unusual, I want to use `ssh-agent` to sign API requests for Chef. It can however be used when authenticating to similarly new SSH servers instead of the original SHA1 method.

https://github.com/whiteley/mixlib-authentication/tree/ssh-agent

whiteley:
@mfazekas any ideas on when the next gem release might be scheduled?

PR #510, opened 2017-03-24
Resolve relative paths in the Include directive properly

knu:
According to [ssh_config(5)](http://man.openbsd.org/ssh_config#Include), relative paths in the Include directive is relative from the directory of the root file, not from that of the including file.  If `~/.ssh/config` includes `~/.ssh/dir1/file1` which has a line `Include dir2/file2`, ssh tries to include `~/.ssh/dir2/file2`, not `~/.ssh/dir1/dir2/file2`.

### DIFF
--- a/lib/net/ssh/config.rb
+++ b/lib/net/ssh/config.rb
@@ -72,9 +72,9 @@ module Net; module SSH
       # ones. Returns a hash containing the OpenSSH options. (See
       # #translate for how to convert the OpenSSH options into Net::SSH
       # options.)
-      def load(path, host, settings={})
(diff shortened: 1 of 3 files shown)


knu:
Here's the implementation: https://github.com/openssh/openssh-portable/blob/61b8ef6a66efaec07e023342cb94a10bdc2254dc/readconf.c#L1492-L1495

I was hit by this incompatibility, which I think should be fixed.

mfazekas:
@knu thanks much for the PR, look good. Can we also add a testcase for it?

knu:
@mfazekas The updated test should cover the change.  It now has a nested include of relative paths.

Issue #513, opened 2017-04-11
Security Issue : Net::Ssh.start load always system default configuation 


fwininger:
### Expected behavior

```
Net::SSH.start(host, user, keys_only: true)
```

expect the documentation behavior :

> :keys_only => set to +true+ to use only private keys from +keys+ and
>   +key_data+ parameters, even if ssh-agent offers more identities. This
>   option is intended for situations where ssh-agent offers many different
>   identites.

### Actual behavior

```
Net::SSH.start(host, user, keys_only: true)
```

load default keys, aka : 

```
# Returns an array of paths to the key files usually defined
# by system default.
def default_keys
  if defined?(OpenSSL::PKey::EC)
    %w(~/.ssh/id_ed25519 ~/.ssh/id_rsa ~/.ssh/id_dsa ~/.ssh/id_ecdsa
      ~/.ssh2/id_ed25519 ~/.ssh2/id_rsa ~/.ssh2/id_dsa ~/.ssh2/id_ecdsa)
  else
    %w(~/.ssh/id_dsa ~/.ssh/id_rsa ~/.ssh2/id_dsa ~/.ssh2/id_rsa)
  end
end
```
### System configuration

- net-ssh 4.1 and the last 5 years versions
- Ruby version 2.4

### Security Issue

This issue allows an hacker to use unknown ssh private key an use it to perform an authentication against a server.

mfazekas:
@fwiniger thanks much for the bugreport

Can you elaborate on the security issue part, so i understand it better?

>This issue allows an hacker to use unknown ssh private key an use it to perform an authentication against a server.


this is ssh config's wording btw whose IdentitiesOnly equivalent is keys_only
```
IdentitiesOnly
             Specifies that ssh(1) should only use the authentication identity
             files configured in the ssh_config files, even if ssh-agent(1) or
             a PKCS11Provider offers more identities.  The argument to this
             keyword must be ``yes'' or ``no''.  This option is intended for
             situations where ssh-agent offers many different identities.  The
             default is ``no''.
```

mfazekas:
I think this is resolved with 5.*

PR #514, opened 2017-04-11
Security Fix : Block the usage of default_keys

fwininger:
According the documentation :

>:keys_only => set to +true+ to use only private keys from +keys+ and
>+key_data+ parameters, even if ssh-agent offers more identities. This
>option is intended for situations where ssh-agent offers many different
>identites.

This fix don't load the default_keys with options[:key_only]

This pull request resolve issues #513 and #489
This pull request resolve pull request #490


mfazekas:
Fixed by #589

fwininger:
Thanks

PR #516, opened 2017-04-26
Support including multiple files by a single Include directive

hanazuki:
ssh_config(5) from OpenSSH says an Include directive may take multiple pathnames. The pathnames are separated by whitespaces and may contain whitespaces when enclosed in double quotes.


mfazekas:
Thanks much for the contribution 👍

PR #517, opened 2017-05-04
add sha256 fingerprints for keys

tmaher:
OpenSSH 6.8 [switched to using SHA256 fingerprints](https://www.openssh.com/txt/release-6.8) a while back. This PR adds SHA256 support to `OpenSSL::PKey::PKey#fingerprint` by adding an optional argument for the algorithm. The default remains MD5.  Unit tests are included for RSA, DSA, and ECDSA NIST curve keys.

If we'd like to more strictly mirror OpenSSH, the MD5 fingerprint format should change. The fixed string "MD5:" should be prepended, e.g. switch from `32:00:44:78:bf:91:02:c1:00:25:0f:f9:0a:f9:aa:c7` to `MD5:32:00:44:78:bf:91:02:c1:00:25:0f:f9:0a:f9:aa:c7`. I held off on that because it'd be a breaking change.

/cc @jameswhite


mfazekas:
Thanks much for the contribution looks great 👍 
We also have `fingerprint` on ed25519 to be changed.

Looking at the code we don't seems to be useing those fingerprints other than logging purposes, do we?

tmaher:
@mfazekas -Whoops, fixed. I've created `net/ssh/pub_key.rb` as a helper module, since the fingerprinting code is the same for both the openssl & ed25519 keys. If you'd rather, I can revert it back to defining `#fingerprint` once each in `openssl.rb` and `ed25519.rb`.

Aside from logging, there's a third key MD5 pubkey fingerprint implementation I haven't touched: see https://github.com/net-ssh/net-ssh/blob/beb299e/lib/net/ssh/transport/kex/diffie_hellman_group1_sha1.rb#L189-L196 . *Note*: all the other kex implementations in that directory are subclasses of `DiffieHellmanGroup1SHA1`, so they're getting the MD5-based key comparisons too. I don't fully understand what that code does yet, but I'm relatively confident none of my changes impact it. It's getting late in my timezone. I can poke at it in a day or two, but I think updating that might better happen in a subsequent PR.

My motivation: my project manages adding/removing public keys to a database on behalf of users. I'd like to use `net-ssh` to both validate the keys are well-formed and to generate key fingerprints for visual reference. I can imagine other people using it for much the same, even if the gem itself only uses `#fingerprint` for logging. That's what I'm more worried about breaking.

mfazekas:
@tmaher thanks looks nice. One thing i'm thinking of is renaming `net/ssh/pub_key` to `net/ssh/authentication/pub_key_fingerprint.rb` also docs should somehow make it clear that the module itself is not part of the API just the fingerprint function. What do you think?

lupine:
We're switching to `net-ssh` to calculate key fingerprints in GitLab: https://gitlab.com/gitlab-org/gitlab-ce/issues/21391 

Currently we use MD5 fingerprints but we'd love to switch to SHA256: https://gitlab.com/gitlab-org/gitlab-ce/issues/20502

Is there anything I can do to help get this PR accepted?

milgner:
Very useful indeed. I'd love to switch our code to SHA256, too!

tmaher:
@mfazekas - I think c53c831 should do what you had in mind re: docs & class hierarchy. The Gemfile changes in c2d6cf9 were necessary to get CI builds for Ruby 2.0 & 2.1 working.

tmaher:
screen shot 2017-11-03 at 1 27 34 pm

nijikon:
@mfazekas I think that we should merge this before we drop support for `ruby < 2.2`.

PR #524, opened 2017-06-08
Rename :paranoid to :verify_host_key

jaredbeck:
The words we choose matter. By using the word "paranoid", we make
fun of people who care about security. Instead, we should encourage
security in our community, or at least use a neutral term.

In this patch, we choose the neutral wording that is used by the SSH
manual page:

> the server must be able to verify the client's host key
> (see the description of /etc/ssh/ssh_known_hosts and
> ~/.ssh/known_hosts, below) for login to be permitted.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,7 @@
+=== Unreleased
+
+ * Deprecate `:paranoid` in favor of `:verify_host_key`
+
 === 4.1.0
 === 4.1.0.rc1
 
--- a/lib/net/ssh.rb
+++ b/lib/net/ssh.rb
@@ -72,7 +72,7 @@ module Net
       :host_name, :user, :properties, :passphrase, :keys_only, :max_pkt_size,
       :max_win_size, :send_env, :use_agent, :number_of_password_prompts,
       :append_all_supported_algorithms, :non_interactive, :password_prompt,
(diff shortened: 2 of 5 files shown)


jaredbeck:
Thanks for the merge, Miklós.

A followup thought: `verify_host_key: :very` is awkward grammar. The other values are fine, but `:very` is a little awkward. I see a few solutions:

1. Do nothing (least disruptive)
2. Deprecate `:very` only, choosing a new word like `:strict`.
3. Deprecate all existing values, replacing them with a set of words that match the classes in the `Net::SSH::Verifiers` module. Values would be replaced as follows:
  - `nil` becomes `:null`, 
  - `true` becomes `:lenient`, 
  - `:very` becomes `:strict`, and
  - `:secure` remains `:secure`.
This would be the most disruptive solution, but perhaps a nice improvement, improving data type consistency (they're all symbols) and expressiveness (they reveal the underlying classes).

If you'd prefer to discuss this followup in a new issue, I will be happy to open one.

mfazekas:
@jaredbeck yeah let's open a new issue for those, thanks!

samueloph:
I would just like to add that i spent a considerable amount of time looking for an option to disable key checking and the setting being called "paranoid" made it more difficult than it should be.

Thanks for changing that :)

PR #525, opened 2017-06-08
Move `none` cipher to end of cipher list

briancain:
If a consumer used the default for the `:encryption` setting and the
host attempted to negotiate cipher suites with only supported suites
after `none` (but supported none), it would go ahead and use `none`
rather than the other supported cipher suites. This commit updates that
behavior to first allow Net::SSH to enumerate all of the possible
default cipher suites before ending on `none`.


mfazekas:
👍  thanks for the contribution!

briancain:
@mfazekas thanks!!

PR #528, opened 2017-06-17
Support multiple identity files

masa0x80:
If configuration files have multiple identity files, Net::SSH should support those.

The [ssh_config manual](http://man.openbsd.org/ssh_config#IdentityFile) states:
> It is possible to have multiple identity files specified in configuration files; all these identities will be tried in sequence. 

We should merge `globals['identityfile']` with `settings['identityfile']` in loading configuration files.

### DIFF
--- a/lib/net/ssh/config.rb
+++ b/lib/net/ssh/config.rb
@@ -144,7 +144,10 @@ module Net; module SSH
           end
         end
 
-        settings = globals.merge(settings) if globals
+        if globals
+          settings['identityfile'] = [globals['identityfile'], settings['identityfile']].flatten.compact
+          settings = globals.merge(settings)
+        end
 
         return settings
       end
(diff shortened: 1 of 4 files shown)


lib/net/ssh/config.rb
```diff
@@ -144,7 +144,10 @@ def load(path, host, settings={}, base_dir = nil)
           end
         end
 
-        settings = globals.merge(settings) if globals
+        if globals
+          settings['identityfile'] = [globals['identityfile'], settings['identityfile']].flatten.compact
```
mfazekas: We could also use merge with block - not sure if it's any cleaner
```
globals.merge(settings) do |key, oldval, newval| 
   case key
     when 'identityfile' 
       oldval + newval
     else
       newval
end
```

mfazekas:
Nice! Thanks for the contribution!

masa0x80:
@mfazekas Thank you for reviewing. I fixed it.

Issue #512, opened 2017-04-04
Wrong debug output with proxy command


imoverclocked:
### Expected behavior

When using a ProxyCommand or similar config the debug output should resemble:

```
D, [2017-04-04T20:50:53.786036 #41101] DEBUG -- net.ssh.transport.session[3fdb40d9d8d8]: establishing connection to proxy command
```

### Actual behavior

Currently, the behavior in debug mode prints a line:

```
D, [2017-04-04T20:50:53.786036 #41101] DEBUG -- net.ssh.transport.session[3fdb40d9d8d8]: establishing connection to some.host:22
```

This disregards proxy configurations where we are not actually connecting to the host:port and can throw debugging off.

https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/transport/session.rb#L64

### Example App

```ruby
gem 'net-ssh', '= 4.0.0.beta3'
require 'net/ssh'
puts Net::SSH::Version::CURRENT

@host = 'localhost'
@user = ENV['USER']
Net::SSH.start(@host, @user, config: 'proxy_command_config') do |ssh|
  puts ssh.exec!('echo "hello"')
end
```

proxy_command_config:
```
Host *
  ProxyCommand nc %h 8080
```

mfazekas:
Thanks for the detailed issue, are you interseted in making a PR solving it?

imoverclocked:
I can't at the moment.

j1wilmot:
I believe this was resolved in #530 and can be closed.

PR #530, opened 2017-06-22
Fix debug output with proxy command

moofkit:
Solving https://github.com/net-ssh/net-ssh/issues/512

Write some tests for this issue.
@imoverclocked, @mfazekas can you guys check is it right?

### DIFF
--- a/lib/net/ssh/transport/session.rb
+++ b/lib/net/ssh/transport/session.rb
@@ -61,12 +61,12 @@ module Net; module SSH; module Transport
       @bind_address = options[:bind_address] || nil
       @options = options
 
-      debug { "establishing connection to #{@host}:#{@port}" }
-
       @socket =
         if (factory = options[:proxy])
+          debug { "establishing connection to #{@host}:#{@port} through proxy" }
           factory.open(@host, @port, options)
         else
+          debug { "establishing connection to #{@host}:#{@port}" }
           Socket.tcp(@host, @port, @bind_address, nil,
                      connect_timeout: options[:timeout])
         end
--- a/test/transport/test_session.rb
+++ b/test/transport/test_session.rb
@@ -1,5 +1,6 @@
 require_relative '../common'
 require 'net/ssh/transport/session'
+require 'net/ssh/proxy/http'
 
 # mocha adds #verify to Object, which throws off the host-key-verifier part of
 # these tests.
@@ -14,9 +15,12 @@ module Transport
   class TestSession < NetSSHTest
     include Net::SSH::Transport::Constants
 
+    TEST_HOST = "net.ssh.test"
+    TEST_PORT = 22
+
     def test_constructor_defaults
-      assert_equal "net.ssh.test", session.host
-      assert_equal 22, session.port
+      assert_equal TEST_HOST, session.host
+      assert_equal TEST_PORT, session.port
       assert_instance_of Net::SSH::Verifiers::Lenient, session.host_key_verifier
     end
 
@@ -48,13 +52,13 @@ module Transport
     def test_host_as_string_should_return_host_and_ip_when_port_is_default
       session!
       socket.stubs(:peer_ip).returns("1.2.3.4")
-      assert_equal "net.ssh.test,1.2.3.4", session.host_as_string
+      assert_equal "#{TEST_HOST},1.2.3.4", session.host_as_string
     end
 
     def test_host_as_string_should_return_host_and_ip_with_port_when_port_is_not_default
       session(port: 1234) # force session to be instantiated
       socket.stubs(:peer_ip).returns("1.2.3.4")
-      assert_equal "[net.ssh.test]:1234,[1.2.3.4]:1234", session.host_as_string
+      assert_equal "[#{TEST_HOST}]:1234,[1.2.3.4]:1234", session.host_as_string
     end
 
     def test_host_as_string_should_return_only_host_when_host_is_ip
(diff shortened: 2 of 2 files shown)


mfazekas:
@moofkit thanks much, looks good to me 👍

moofkit:
ok! but we can upgrade this feature. i.e. if it is http proxy then message would be `establishing connection to foo.bar:22 with http proxy`; with command proxy `establishing connection to foo.bar:22 with command proxy` or smth like that.

moofkit:
Ok, it's ready for review.

I've implemented simple version. If it will be good to know the type of proxy in the debug log, then I can do it too.

lib/net/ssh/transport/session.rb
```diff
@@ -61,12 +61,12 @@ def initialize(host, options={})
       @bind_address = options[:bind_address] || nil
       @options = options
 
-      debug { "establishing connection to #{@host}:#{@port}" }
-
       @socket =
         if (factory = options[:proxy])
+          debug { "establishing connection to proxy" }
```
mfazekas: wouldn't make sense to print host port too? like `proxy #{@host}:#{@port}`? I guess any more clarification could be done by `open` method in proxy.

imoverclocked: Nope, it doesn't make sense to place `@host` or `@port` because the proxy config ignores those settings entirely. 

If I have an SSH config that says "when I connect to foo.bar.com, proxy me baz.quux.org" then this line would state "connecting to foo.bar.com" even though there is no connection actually being made to "foo.bar.com."

imoverclocked: If we wanted to be pedantic here we would say:

"establishing proxy connection for #{host}"

and then ensure the proxy connection we are opening has a similar level of debugging so that we would eventually see:

"establishing connection to baz.quux.org"

However, as it stands, this PR is a big improvement over what exists today.

mfazekas: > Nope, it doesn't make sense to place @host or @port because the proxy config ignores those settings entirely.

Does it ignores it? It does connect to there but through a proxy doesn't it?  What about something like:

```
establishing connection to #{@host}:#{@port} through proxy
```

imoverclocked: It is completely ignored. When connecting through a proxy, you only have one connection: the proxy. Once you have established a proxy connection and make a request akin to `ssh -W host:port proxy.host.name` the proxy connects on your behalf to *something* which might (or might not) be host:port.

In any case, the only connection the client makes is to a proxy. Debugging here should report connecting to a proxy because that's the connection being made. It's also the connection that will fail in this context. We need to inform the user that it's the connection to the proxy that failed, not the connection to the final endpoint.

mfazekas: > It is completely ignored
Not sure what do you mean it's ignored, it's not ignored for sure. For example in the ssh command example you showed: `ssh -W host:port proxy.host.name` the command does connect to `host:port` through the intermediate host `porxy.host.name`, you might be connecting to `host:port` indirectly, but you're still connecting there (through a proxy).

The connection can fail if the proxy is not available or it can fail or the host/port you're connecting to through the proxy is not available.

imoverclocked: Ok, deeper dive on the happy path:

1. Make TCP connection to proxy (start of the factory call)
1. SSH auth to proxy (what successive debug messages from the factory may talk about)
1. Open channel to proxy over established connection 
1. Send direct-tcpip request (https://tools.ietf.org/html/rfc4254#section-7.2)
  a. proxy attempts to connect to some endpoint
  b. proxy replies with, "ok, this channel now represents a connection to remote port"
  c. (factory returns channel as if it's a socket)
1. SSH auth + requests to the destination host

When attempting to debug interactions with a proxy, it's actually step 4 that is making a connection to the destination host. Given that steps 1,2,3 can all fail before step 4 is attempted, simply stating "connecting to #{@host}:#{@port} through proxy" here is somewhat misleading.

In the best scenario, we might consider passing debug level flags into the factory so that we can illuminate steps 1-4 in which case this debug message goes away entirely.

Like I said earlier, this message (and really any variation that anyone have mentioned) are a vast improvement over current behavior.

If we want to pull from OpenSSH logging behavior as a model, then:

```
$ ssh -v -o ProxyCommand="ssh -W %h:1 myproxy.example.com" destination.example.com
OpenSSH_7.2p2, LibreSSL 2.4.1
debug1: Reading configuration data /Users/imoverclocked/.ssh/config
debug1: /Users/imoverclocked/.ssh/config line 38: Applying options for *
debug1: Reading configuration data /etc/ssh/ssh_config
debug1: /etc/ssh/ssh_config line 20: Applying options for *
debug1: Executing proxy command: exec ssh -W destination.example.com:1 myproxy.example.com
...
```

```
$ ssh -v -W destination.example.com:1 myproxy.example.com
OpenSSH_7.2p2, LibreSSL 2.4.1
debug1: Reading configuration data /Users/imoverclocked/.ssh/config
debug1: /Users/imoverclocked/.ssh/config line 38: Applying options for *
debug1: Reading configuration data /etc/ssh/ssh_config
debug1: /etc/ssh/ssh_config line 20: Applying options for *
debug1: Connecting to myproxy.example.com [1.1.1.1] port 22.
debug1: fd 3 clearing O_NONBLOCK
debug1: Connection established.
... auth stuff ...
debug1: channel_connect_stdio_fwd destination.example.com:1
debug1: channel 0: new [stdio-forward]
debug1: getpeername failed: Bad file descriptor
... connection shutdown ...
```

ie: the top level ssh process mentions that it is starting a proxy command and it's up to the proxy (or proxy command in this case) to talk about connecting to destination.example.com:1

mfazekas: @imoverclocked i appreciate all the comments, but i'm sure we both can spend our times better than arguing on this any further.

I prefer the message to include host and port eg:
```
establishing connection to #{@host}:#{@port} through proxy
```

mfazekas: Please change it to
```
establishing connection to #{@host}:#{@port} through proxy
```

imoverclocked: 👍 Thanks for your great work!

PR #543, opened 2017-08-14
Add key to agent functionality

montdidier:
This PR covers the desired use case for a deployment scenario I wanted to use. It may be that others want this functionality also.

### DIFF
--- a/lib/net/ssh.rb
+++ b/lib/net/ssh.rb
@@ -63,7 +63,7 @@ module Net
     # This is the set of options that Net::SSH.start recognizes. See
     # Net::SSH.start for a description of each option.
     VALID_OPTIONS = [
-      :auth_methods, :bind_address, :compression, :compression_level, :config,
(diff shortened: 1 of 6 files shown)


lib/net/ssh.rb
```diff
@@ -103,6 +103,7 @@ module SSH
     #
     # This method accepts the following options (all are optional):
     #
+    # * :add_keys_to_agent => yes, no, confirm
```
mfazekas: i assume this would be true/false, :confirm, the code bellow does not supports the `:confirm` option does it?

mfazekas:
Thanks much for the PR and sorry for the late answer. 
Assume this is equivalent to AddKeysToAgent from http://www.openssh.com/txt/release-7.2 we either need to add confirm support or document that its not supported. Not sure how to translate from config files in that case, probably to false

mfazekas:

PR #544, opened 2017-09-07
Remove Net::SSH::Compat.io_select

milgner:
Despite the comment, it was not doing anything anymore except invoking the default `IO.select`.

### DIFF
--- a/lib/net/ssh/buffered_io.rb
+++ b/lib/net/ssh/buffered_io.rb
@@ -113,7 +113,7 @@ module Net; module SSH
     def wait_for_pending_sends
       send_pending
       while output.length > 0
-        result = Net::SSH::Compat.io_select(nil, [self]) or next
+        result = IO.select(nil, [self]) or next
         next unless result[1].any?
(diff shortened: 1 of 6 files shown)


mfazekas:
I think @HoneyryderChuck might have been using that to monkey patch net-ssh io_select to use celluloid.

milgner:
Oh... I see... 😄 In that case I'd opt for just updating the comment. Right now it looked like it was just left-over code. I was investigating a weird connection problem (turned out to be something else though) and when I found this I thought I'd just quickly help out and clean this up... 😬

HoneyryderChuck:
@mfazekas , I'm not working on that project anymore, so the answer is "maybe" :) .

If I were you, I'd accept this change. It cleans up internals, and it's not really public API, i.e. I wouldn't support monkey-patches based on it. If the current maintainers of the project notice any hiccup, they'll either not upgrade or proceed to find a workaround, or maybe work with the current event loop API, and you will hear from them.

PR #545, opened 2017-09-18
use @prompt from the base class

busterb:
It looks like when 2b6cd7a5a139ce2c1379546b1e8d09c9013ebbd2 refactored the prompt methods, the @prompt instance variable from the base class does not get referenced correctly.

Noted here https://github.com/rapid7/metasploit-framework/issues/8970#issuecomment-329937858

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,12 +7,9 @@ addon:
     gateway.netssh
 
 rvm:
-  - 2.0
-  - 2.1
   - 2.2
-  - 2.3.0
-  - 2.4.0
-  - jruby-9.1.6.0
+  - 2.3.5
+  - 2.4.2
   - rbx-3.69
   - ruby-head
 env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,23 +10,23 @@ rvm:
   - 2.2
   - 2.3.5
   - 2.4.2
-  - rbx-3.69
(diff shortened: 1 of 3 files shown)


busterb:
Note, this may break some of the tests that rely on defining prompt. I wonder if maybe we're using the library incorrectly...

mfazekas:
Hi thanks for the PR, i'm curious what's the issue you see?

There is a
`attr_reader :prompt`
in abstract, so it's fine to use `prompt` to access the instance variable vs `@prompt`

busterb:
The test case that we are looking at is

```
require 'net/ssh'

ip = '192.168.1.181'
opt_hash = {
  port: 22,
  auth_methods: ['password', 'keyboard-interactive'],
  use_agent: false,
  config: false
}

transport = Net::SSH::Transport::Session.new(ip, opt_hash)

auth = Net::SSH::Authentication::Session.new(transport, opt_hash)
auth.authenticate('ssh-connection', 'root', 'passwordthatdoesntexist')

puts "#{ip} connection good -- #{auth.transport.server_version.version}"
```

On failure, authenticate raises an exception that prompt is Nil: https://github.com/rapid7/metasploit-framework/issues/8970#issuecomment-329935969

I'll close this out for now, since it doesn't solve the problem.

Issue #374, opened 2016-05-18
Net::SSH does not understand Match blocks


djwmarks:
I have an SSH configuration including a Match Host block, like the following:

```
Match Host *.foobar.com
  ProxyCommand command-to-run
```

The ProxyCommand is only valid within the Match Host context, yet it is configured for every host.

```
irb(main):001:0> require "net/ssh"
=> true
irb(main):002:0> Net::SSH::Config.for("127.0.0.1")
=> {:auth_methods=>[...], :proxy=>#<Net::SSH::Proxy::Command:0x007f9c0bb5c1e8 @command_line_template="...", @command_line=nil>, :user=>"...", :keys=>["..."], :send_env=>[/^LANG$/, /^LC_.*$/]}
```

I could maintain separate configuration files but that rather defeats the point of having a configuration parser. Of course the ideal would be that Net::SSH understands the complete syntax. However, if Net::SSH can't understand the full gamut of SSH configuration options, perhaps you could add a methods of blacklisting specific lines as a compromise? Something akin to Rubocop directives, e.g.:

```
# Net::SSH::Config ignore
Match Host *.foobar.com
  ProxyCommand command-to-run
# Net::SSH::Config clear
```

This would give people a workaround for future syntax changes as well while the parser catches up.

Cheers

djwmarks:
PR https://github.com/net-ssh/net-ssh/pull/413

mfazekas:
See also #593

PR #553, opened 2017-10-25
Add initial Match support

hanazuki:
issue: https://github.com/net-ssh/net-ssh/issues/374

This change makes Net::SSH::Config to recognize `Match` directive in ssh_config.
Only the `all` criteria keyword (and its negation) is supported at this time.

Though only supporting `all` is not very useful, it is important that net-ssh correctly handles block structure of ssh_config.


mfazekas:
Thanks much for the PR! One the CI fixed on master this is good to go! 
👍

hanazuki:
Merged master to this PR to check the CI works.

PR #554, opened 2017-11-25
Support multiple config values

stefansundin:
... for LocalForward, RemoteForward and DynamicForward.

I realize that net-ssh does not automatically use these keys, but I use net-ssh to load these values into my own program.

Here's another issue I've had recently, that I can open a PR to fix if you agree:

With the following config:
```
Host test1.example.com
  DynamicForward 123

Host test2.example.com
  DynamicForward localhost:123
```

Then running the following:
```
irb(main):002:0> Net::SSH::Config.load("~/.ssh/config", "test1.example.com")
=> {"host"=>"test1.example.com", "dynamicforward"=>123}
irb(main):003:0> Net::SSH::Config.load("~/.ssh/config", "test2.example.com")
=> {"host"=>"test2.example.com", "dynamicforward"=>"localhost:123"}
```

I now have two different types for the same option, the first is an integer and the second is a string. This is a bit confusing. In my opinion the type should depend on the key name only. It makes sense to coerce the port number (are there any other option?). The side effect is that I have to call `.to_s` to make sure it's a string before processing it.

Another problem with this overly aggressive type coercion is the following:

```
Host test3.example.com
  User yes
```

```
irb(main):004:0> Net::SSH::Config.load("~/.ssh/config", "test3.example.com")
=> {"host"=>"test3.example.com", "user"=>true}
```

Although unlikely to happen, this is obviously incorrect.

Looking quickly, it appears that openssh is determining the types based on what key it is. https://github.com/openssh/openssh-portable/blob/b7548b12a6b2b4abf4d057192c353147e0abba08/readconf.c#L876-L987

Let me know what you think. Thanks!!

### DIFF
--- a/lib/net/ssh/config.rb
+++ b/lib/net/ssh/config.rb
@@ -46,6 +46,9 @@ module Net; module SSH
       # it's used by some clients to query the server for allowed auth methods
       @@default_auth_methods = %w(none publickey password keyboard-interactive)
 
+      # The following options can be repeated
+      @@multi_options = %w(identityfile localforward remoteforward dynamicforward)
+
       # Returns an array of locations of OpenSSH configuration files
       # to parse by default.
       def default_files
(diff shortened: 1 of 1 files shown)


stefansundin:
I rebased this PR on the latest master. I am getting an appveyor error saying the method is too complex now, but I think other parts of it are better suited to be refactored. If you'd like, I can wrap my other ideas (from the PR description above) into this PR and hopefully make the test pass.

By the way, does this line have a typo? https://github.com/net-ssh/net-ssh/blob/9f2384dacd281b3bb353c82444501fdfba098dc0/lib/net/ssh/config.rb#L134

Shouldn't it be `unless globals.key?(key)`?

mfazekas:
@stefansundin sorry for the late answer, i think the PR is nice, but some unittest would be needed.

I'm not sure that `globals[key] = value unless settings.key?(key) ` is a typo, as the docs states that existings settings should take a precedence, on the other hand we do that later, in the globals.merge(settings),  
So the difference `globals[key] = value unless settings.key?(key) ` and `globals[key] = value unless globals.key?(key)` is that if we take the first or last value for that key. Not sure what is the correct.

mfazekas:

PR #555, opened 2017-11-26
rbnacl 5.x

nijikon:


nijikon:
Ping

mfazekas:
What is the reason for dropping old ruby version? As such change is binary incompatible we can do it with a major version change.

nijikon:
@mfazekas rbnacl-5.0.0 requires ruby version >= 2.2.6

mfazekas:
But net-ssh works without rbnacl and with older rbnacl, so there is probably no need to drop old rubies

lib/net/ssh/authentication/ed25519.rb
```diff
@@ -1,4 +1,4 @@
-gem 'rbnacl', '>= 3.2.0', '< 5.0'
+gem 'rbnacl', '>= 3.2.0'
```
mfazekas: i'm not sure if we want to allow any future rbnacl, in theory the might not be compatible, on the otherhand they'll be likely compatible for what we use.

nijikon:
I think that we should not encourage people to use ruby versions that are not supported by the ruby-core team any longer and does not receive any fixes.

lib/net/ssh/authentication/ed25519.rb
nijikon: If it works I don't see how we should freeze it; maybe we can block it `'>= 3.2.0', '< 6.0'`. What do you think?

mfazekas:
> I think that we should not encourage people to use ruby versions that are not supported by the ruby-core team any longer and does not receive any fixes.

It's true, but we cant drop ruby releases except in major revisions as we try to follow SemVer.

nijikon:
I agree, so can we bump the major revision?

nijikon:
I can extract ruby version drop to a separate PR if that helps.

mfazekas:
> I can extract ruby version drop to a separate PR if that helps.

Yes please! I usually drop ruby versions when we release a new major version, not the other way around. I think relaxing rbnacl version req and dropping old ruby releases are separate issues

PR #560, opened 2017-12-03
drop support for ruby lower than 2.2.6

nijikon:
Drop ruby versions that are not supported by the ruby-core team any longer and does not receive any fixes.


nijikon:
@mfazekas can we merge this?

PR #563, opened 2017-12-15
Replace RbNaCl dependency with ed25519 gem

tarcieri:
RbNaCl is a heavyweight dependency that is only being used for Ed25519 signatures. The ed25519 gem is purpose-made for ed25519 signing and verification and should install much more quickly:

https://github.com/cryptosphere/ed25519

This change otherwise preserves ed25519 being used as an optional dependency and generally tries to keep as much as possible the same.


tarcieri:
I'm the author of both RbNaCl and the ed25519 gem. It looks like you've been struggling with RbNaCl a bit, and I don't blame you, which is why I made this gem.

It's actually 5 years old and predates RbNaCl, but I've recently dusted it off and upgraded the underlying native implementations it uses. It provides a C extension backend based on the same implementation as libsodium, and also a fast Java implementation on JRuby.

Gemfile
```diff
@@ -7,10 +7,6 @@ if !Gem.win_platform? && RUBY_ENGINE == "ruby"
   gem 'byebug', group: [:development, :test]
 end
 
-if (Gem::Version.new(RUBY_VERSION) <=> Gem::Version.new("2.2.6")) < 0
-  gem 'rbnacl', '< 4.0'
-end
```
tarcieri: Wasn't sure what this was about specifically. Let me know if it needs to be preserved.

Gemfile.norbnacl
```diff
@@ -1,10 +0,0 @@
-source 'https://rubygems.org'
```
tarcieri: Wasn't sure what this was all about, but hopefully I can address whatever problems were arising which brought about the need for this in the first place in the ed25519 gem.

lib/net/ssh/authentication/ed25519.rb
```diff
@@ -19,18 +10,25 @@
 
 module Net; module SSH; module Authentication
 module ED25519
-  class SigningKeyFromFile < RbNaCl::Signatures::Ed25519::SigningKey
```
tarcieri: This seemed... kind of crazysauce to me. Using a `SimpleDelegator` should preserve the semantics (the tests seem to be passing) while decoupling it from the internal implementation details of the `Ed25519::SigningKey` class.

net-ssh.gemspec
```diff
@@ -28,9 +28,8 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
-  unless ENV['NET_SSH_NO_RBNACL']
-    spec.add_development_dependency("rbnacl-libsodium", "~> 1.0.11")
-    spec.add_development_dependency("rbnacl", ['>= 3.2.0','< 5.0'])
+  unless ENV['NET_SSH_NO_ED25519']
```
tarcieri: I renamed this environment variable as we're no longer using RbNaCl, so heads up.

This environment variable name probably makes more sense anyway.

lib/net/ssh/authentication/ed25519.rb
```diff
@@ -19,18 +10,25 @@
...
-      @signing_key = sk
-      @verify_key = RbNaCl::Signatures::Ed25519::VerifyKey.new(pk)
+      key = ::Ed25519::SigningKey.from_keypair(sk)
+      raise ArgumentError, "pk does not match sk" unless pk == key.verify_key.to_bytes
+
+      super(key)
+    end
+
+    # Deprecated method provided by the RbNaCl API
+    def keypair_bytes
```
tarcieri: This should be the only method available in the RbNaCl implementation which is not available in the ed25519 gem. I went ahead and put this here for compatibility but changed all of the usages of it throughout the gem to the new API (which is just `#keypair` and is an attr reader)

Gemfile.norbnacl
mfazekas: The idea about .norbnacl Gemfile is that we could ran test with rbnacl and without rbnacl.
It was required on windows as we did not have rbnacl there.

Gemfile
mfazekas: I think rbnacl 4 required ruby 2.2.6 so on earlier ruby-s we've installed rbnacl 3

lib/net/ssh/authentication/ed25519.rb
mfazekas: Yeah i'm not sure what was my point here either. This is not really part of the api and we can release a new version so we can get a rid of it.

mfazekas:
Thanks much for the PR, looks good! I think i'll need to bump major version.
I think the biggest issue with RbNacl was windows support. I assume ed25519 doesn't have precompiled binaries either, but likely will be easier to get it compile on windows.

tarcieri:
@mfazekas rbnacl-libsodium now has a build script that, in theory, will allow me to cross-compile and release Windows binary gems.

I can try to do the same with the ed25519 gem. Looks like it's passing on Appveyor though?

dentarg:
> The ed25519 gem is purpose-made for ed25519 signing and verification and should install much more quickly

This would be really nice :)

mfazekas:
Thanks much for the PR! It's now merged via #564 

One  hope with libsodium was that we can support other algs with libsodium. Unfortunately the openssh and libsoidum implements different variants, and this did not work out in case of [chacha20-poly1305@openssh.com](https://cvsweb.openbsd.org/cgi-bin/cvsweb/src/usr.bin/ssh/PROTOCOL.chacha20poly1305?annotate=HEAD) for example which has different [variants](https://download.libsodium.org/doc/secret-key_cryptography/chacha20-poly1305.html) implemented in libsodium. The `ed25519` gem should simplify installation, thanks much!

tarcieri:
Awesome, thanks!

mfazekas:
@tarcieri FWIW we're back at RbNaCl for `poly1305`. See #908

Issue #542, opened 2017-07-27
Slow file upload via CTR mode of encryption


gsecillano:
### Expected behavior

Uploading via SCP using CTR encryption mode should be comparable to using CBC.

### Actual behavior

CTR encryption mode is ~46x slower than CBC.  For a 79MB file upload via SCP:

*aes128-ctr*
```
real    1m17.345s
user    1m15.700s
sys     0m1.464s
```
*aes128-cbc*
```
real    0m2.538s
user    0m1.688s
sys     0m0.680s
```

### System configuration

- net-ssh version: 2.7.0 (reproducible on 4.1.0)
- Ruby version: 2.2.6 (jemalloc)

### Example App

```ruby
#gem 'net-ssh', '= 2.7.0'
gem 'net-scp', '= 1.1.2'
require 'net/scp'
puts Net::SSH::Version::CURRENT

@host = ENV['HOST']
@user = ENV['USER']
Net::SCP.start(@host, @user, encryption: ENV['ENCRYPTION'] || 'aes128-cbc') do |scp|
  # The file /tmp/testdata was created using
  #   dd if=/dev/random of=/tmp/testdata count=1024000
  scp.upload!('/tmp/testdata', '/tmp/testdata')
end
```

mfazekas:
@gsecillano can you please try #570  and #569

PR #569, opened 2018-01-27
improve CTR speed by not mutating the remaining string in a loop

apachelogger:
#slice! in a loop gets expensive quickly when encrypting larger data blobs
(e.g. when uploading data through net-sftp).
#slice! needs to modify self AND create a new string for the substr,
#slice on the other hand only needs to do the latter.

as we do not need the modified remaining string in the loop we can simply
slice in the loop and then change remaining afterwards to modify the
string permanently.

Closes: #542

### DIFF
--- a/lib/net/ssh/transport/ctr.rb
+++ b/lib/net/ssh/transport/ctr.rb
@@ -12,7 +12,7 @@ module Net::SSH::Transport
         @counter_len = orig.block_size
         orig.encrypt
         orig.padding = 0
-        
+
         singleton_class.send(:alias_method, :_update, :update)
         singleton_class.send(:private, :_update)
         singleton_class.send(:undef_method, :update)
@@ -50,11 +50,14 @@ module Net::SSH::Transport
 
           encrypted = ""
 
-          while @remaining.bytesize >= block_size
-            encrypted += xor!(@remaining.slice!(0, block_size),
+          offset = 0
+          while (@remaining.bytesize - offset) >= block_size
+            encrypted += xor!(@remaining.slice(offset, block_size),
                               _update(@counter))
             increment_counter!
+            offset += block_size
           end
+          @remaining = @remaining.slice(offset..-1)
 
           encrypted
         end


apachelogger:
FWIW, the entire CTR implementation is a major bottleneck, when using net-sftp to upload files at least half the execution time is spent *only* in the ctr implementation (that is CTR-only, excluding the underlying cipher!). I expect much of the net-sftp performance whining you can find on the web is in fact because the CTR is so slow.

It may well be much more efficient to use the existing `*-ctr` ciphers in openssl when available (which should be on just about all current linux systems) instead of using this ruby CTR implementation. openssh also uses openssl instead of their own CTR implementation, when available. I tried to make that change but failed horribly as I am lazy and haven't read the RFCs, so my understand of SSH packet building and encryption is fairly limited :sob:

lib/net/ssh/transport/ctr.rb
```diff
@@ -50,11 +50,16 @@ def update(data)
...
+          offset = 0
+          while (@remaining.bytesize - offset) >= block_size
+            encrypted += xor!(@remaining.slice(offset, block_size),
                               _update(@counter))
             increment_counter!
+            offset += block_size
           end
+          # only modify remaining after loop, we do not need it modified while
+          # looping
+          @remaining = @remaining.slice(offset..-1)
```
mfazekas: I'd prefer `slice!` here unless this version is there is a measureable perf improvement. Also I don't think the comment above is neccesary the code itself is readable.

mfazekas:
Thanks much for the PR, looks great!

Do you have a benchmark, for comparing the old/new?
Sure using openssl-s impls would be great a probably much much faster.

lib/net/ssh/transport/ctr.rb
mfazekas: yep, you're right i've overlooked that `slice!` has different return than `slice`

apachelogger:
When doing 16 `update` calls the slice change is 17% faster overall (49s vs. 40s) on a bytesize of 32800 and a block_size of 16. That doesn't seem too huge, but ...

In practise that translates to a speed increase from 2.6 MiB/s to 4.2 MiB/s (with 2 pending requests and slight monitoring overhead)!

`#slice!` uploading 1.5GiB to localhost
`42,02s user 0,32s system 99% cpu 42,718 total`

`#slice` uploading 1.5GiB to localhost
`25,70s user 0,31s system 98% cpu 26,427 total`

(openssh's sftp transfers with `3,65s user 1,21s system 65% cpu 7,407 total`, so there's still plenty of room for improvement :wink:)

mfazekas:
@apachelogger thanks for the numbers it would be still great to have a reproducible script that can demonstrate the speedup.

apachelogger:
I am not sure what you are imagining exactly. But this is what I use for testing 

```
require 'net/sftp'

class Monitor
  def on_open(uploader, file)
    puts "starting upload: #{file.local} -> #{file.remote} (#{file.size} bytes)"
    @start = Time.now
  end

  def on_put(uploader, file, offset, data)
    passed = Time.now - @start
    average = offset / passed
    puts "writing #{data.length} bytes to #{file.remote} starting at #{offset} in #{average / 1024 / 1024} mb/s"
  end
end

Net::SFTP.start('localhost', 'me') do |sftp|
  file = 'test.iso'
  STDERR.puts "Uploading #{file}..."
  sftp.upload!(file, '/tmp/test.iso', read_size: 32_768, requests: 2, progress: Monitor.new)
end
```

PR #575, opened 2018-02-09
Ensure functionality in FIPS enabled systems

### LINKED ISSUE #574: net-ssh should avoid non-FIPS compliant cryptography when used on a FIPS-enabled system
### Expected behavior

When using net-ssh on a system that has been set in`fips` mode, the code should not crash at any point.

### Actual behavior

Presently, the use of `OpenSSL::Digest::MD5.hexdigest` will cause an immediate failure on a FIPS-enabled system.

### System configuration

- net-ssh version: 4.2.0
- Ruby version: 2.4.1 and 2.1.9

### Example App

Any use of `Net::SSH.start` will cause the crash. The easiest method for reproducing the issue is to boot a CentOS 7 system in `fips` mode and then run a simple `Net::SSH.start` against any system using the **system** Ruby.

Full instructions can be found [in the Red Hat Documentation](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/security_guide/chap-federal_standards_and_regulations)

trevor-vaughan:
FIPS enabled system are restricted in which algorithms may be used.
Without restricting the underlying cryptographic methods used by
net/ssh, applications built around the library may crash in unexpected
ways during use.

This patch detects whether or not non-FIPS algorithms can be used and
adjusts the default settiings accordingly.

Additionally transport/algorithms was updated to print the client and
server algorithms that are in use so that users can easily detect a
server/client algorithm mismatch in the future.

Closes #574

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,6 +46,8 @@ install:
 
 script:
   - ssh -V
+  - sudo chown -R travis:travis /home/travis/.ssh
+  - chmod -R go-rwx /home/travis/.ssh
   - bundle _1.16_ exec rake test
   - BUNDLE_GEMFILE=./Gemfile.noed25519 bundle _1.16_ exec rake test
   - bundle _1.16_ exec rake test_test
--- a/lib/net/ssh.rb
+++ b/lib/net/ssh.rb
@@ -76,6 +76,21 @@ module Net
       :agent_socket_factory, :minimum_dh_bits, :verify_host_key
     ]
 
+    # Provides a fingerprint operation on the passed blob
+    #
+    # Adjusts for FIPS and non-FIPS systems as appropriate
+    def self.fingerprint(blob)
+      unless blob.is_a?(String)
+        raise ArgumentError, "item to fingerprint must be a String"
+      end
+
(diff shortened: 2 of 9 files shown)


mfazekas:
For fingerprints we probably want to do what openssh-does there is a config variable FingerprintHash and we should also detect the variant prefix. 

https://www.openssh.com/txt/release-6.8

trevor-vaughan:
@mfazekas Interesting. I'll mod the PR to use `base64` when creating a fingerprint and using SHA256 later tonight. (Allow edits from maintainers is on if anyone wants to get to it before then).

Unfortunately, I don't know of a foolproof method for detecting FIPS mode except for watching MD5 crash and burn. I keep hoping that the ruby `openssl` project will eventually have a detection flag but it doesn't seem to at this point.

mfazekas:
@trevor-vaughan can you please check the master branch, fingerprinting is now customizable and some places it defaults to 'SHA256' see #591

trevor-vaughan:
@mfazekas That looks like it should do what I need. Thanks for the heads up!

Issue #110, opened 2013-07-01
ruby-net-ssh: can't add a new key into hash during iteration during ssh.exec


martin21:
I am currently packaging our own distkeys key distribution ruby script
(see http://bugs.debian.org/712787 RFS: distkeys/1.0 -- distribute SSH keys).

However it only works with Ruby 1.8 for now, as with Ruby 1.9 I get a error
back from ruby-net-ssh when trying to add or remove a key:

```
./distkeys -K somekey.pub -H somehost add

Host: somehost
Connecting to host somehost (user: ms, port: 9999)...
Opening SFTP session...
Key somekey added.
Creating a backup to .ssh/authorized_keys-2013-07-01.bak if not already done today...
Uploading keys to .ssh/authorized_keys-new...
File does exist and has correct size, moving to .ssh/authorized_keys...
/var/lib/gems/1.9.1/gems/net-ssh-2.6.7/lib/net/ssh/connection/session.rb:299:in `[]=': can't add a new key into hash during iteration (RuntimeError)
        from /var/lib/gems/1.9.1/gems/net-ssh-2.6.7/lib/net/ssh/connection/session.rb:299:in `open_channel'
        from /var/lib/gems/1.9.1/gems/net-ssh-2.6.7/lib/net/ssh/connection/session.rb:323:in `exec'
        from /var/lib/gems/1.9.1/gems/net-ssh-2.6.7/lib/net/ssh/connection/session.rb:357:in `exec!'
        from ./distkeys:206:in `block in commit'
        from /var/lib/gems/1.9.1/gems/net-sftp-2.1.2/lib/net/sftp/request.rb:87:in `call'
        from /var/lib/gems/1.9.1/gems/net-sftp-2.1.2/lib/net/sftp/request.rb:87:in `respond_to'
        from /var/lib/gems/1.9.1/gems/net-sftp-2.1.2/lib/net/sftp/session.rb:948:in `dispatch_request'
        from /var/lib/gems/1.9.1/gems/net-sftp-2.1.2/lib/net/sftp/session.rb:911:in `when_channel_polled'
        from /var/lib/gems/1.9.1/gems/net-ssh-2.6.7/lib/net/ssh/connection/channel.rb:311:in `call'
        from /var/lib/gems/1.9.1/gems/net-ssh-2.6.7/lib/net/ssh/connection/channel.rb:311:in `process'
[...]

PR #594, opened 2018-03-23
Fix SSH::Connection::Session#busy? iterating channel hash

2called-chaos:
Maybe I'm using it wrong (I assumed some thread safety) but if I were to loop the SSH connection in a thread and creating new channels in another thread ([my code that breaks](https://github.com/2called-chaos/db_sucker/blob/master/lib/db_sucker/application/dispatch.rb#L121-L139)) I sometimes get:

```
RuntimeError: can't add a new key into hash during iteration
.../net-ssh-4.2.0/lib/net/ssh/connection/session.rb:344:in `open_channel'
```

This PR dups the channel hash in the `SSH::Connection::Session#busy?` method which caused my issue (creating a new channel while busy? is iterating over it by using `any?` with a block and params).

I'm not sure if this creates other problems, e.g. busy returns false with the dupped hash while there was a new one inserted into the real one.

---

This is kinda related to #110 I suppose.

There are a few more places where `channels` are iterated without using the `each_channel` method (which dups the hash) but I'm not sure if there are issue with using dup or #110 just forgot them.

* https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/connection/session.rb#L125
* https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/connection/session.rb#L563

---

Additional info:

* Why is there no need to dup the other `any?` in the `busy?` method? The reason is the C implementation, it only iterates if there is a block given.
* This is only an issue if you were to use `ssh.loop` without a custom block (or use `busy?(false)` in it), this is why I haven't noticed that earlier since my main application uses a [custom check](https://github.com/2called-chaos/db_sucker/blob/master/lib/db_sucker/application/sklaven_treiber.rb#L170-L174) to see if there are pending channels.

### DIFF
--- a/lib/net/ssh/connection/session.rb
+++ b/lib/net/ssh/connection/session.rb
@@ -155,7 +155,7 @@ module Net
           if include_invisible
             channels.any?
           else
-            channels.any? { |id, ch| !ch[:invisible] }
+            channels.dup.any? { |id, ch| !ch[:invisible] }
           end
         end
     


mfazekas:
@2called-chaos #110 isue was that net-ssh was not reentrant, but net-ssh is not thread safe. 

Sorry but i don't like the idea of adding `dup` at places. I'd recommend using locks or other mechanism to make sure you don't call net-ssh methods from different threads at the same time.

Issue #532, opened 2017-06-26
Awkward grammar: `verify_host_key: :very`


jaredbeck:
After renaming `:paranoid` to `:verify_host_key` (https://github.com/net-ssh/net-ssh/pull/524) it becomes apparent that `verify_host_key: :very` is awkward grammar. The other values are fine, but `:very` is awkward. I see a few solutions:

1. Do nothing (least disruptive)
2. Deprecate `:very` only, choosing a new word like `:strict`.
3. Deprecate all existing values, replacing them with a set of words that match the classes in the `Net::SSH::Verifiers` module. Values would be replaced as follows:
        - `nil` becomes `:null`, 
        - `true` becomes `:lenient`, 
        - `:very` becomes `:strict`, and
        - `:secure` remains `:secure`.
  This would be the most disruptive solution, but perhaps a nice improvement, improving data type consistency (they're all symbols) and expressiveness (they reveal the underlying classes).

I'm happy with all of these options, including doing nothing, but I thought I'd mention it anyway.

mfazekas:
@jaredbeck Thanks, we're open to a PR on this. I'm ok with either 2 or 3. 
I'm closing this now, will reopen on further input.

jaredbeck:
Thanks Miklós. I will open a PR implementing solution 3 (all symbols).

PR #595, opened 2018-03-24
New options for :verify_host_key

jaredbeck:
Deprecate some existing values, replacing them with a set of words
that match the classes in the `Net::SSH::Verifiers` module. 

<strike>Values would be replaced as follows:

- `false` becomes `:null`,
- `true` becomes `:lenient`,
- `:very` becomes `:strict`, and
- `:secure` remains `:secure`.
</strike>

[Edit: See discussion below for final names chosen.]

This is a nice improvement, improving data type consistency (they're
all symbols) and expressiveness (they reveal the underlying classes).

This change was preliminarily approved in
https://github.com/net-ssh/net-ssh/issues/532


jaredbeck:
Travis passes, appveyor fails with:

```
Authentication::TestPageapnt#test_agent_should_raise_without_pagent [C:/projects/net-ssh/test/win_integration/test_pageant.rb:26]:
[Net::SSH::Authentication::AgentNotAvailable] exception expected, not Class: <Net::SSH::Exception>
```

This failure seems unrelated to my changes.

mfazekas:
@jaredbeck thanks for the change. 

I'm fine with the proposed terminology but if we can improve that would be even better.
I'm thinking on changing the terminology like:

1. strict:
`verify_host_key: :accept_new` reads better than `verify_host_key: :strict` and strict is also a bit confusing because StrictHostKeyChecking=yes in openssh is our secure option. What about `verify_host_key: :except_new`?

2. lenient:
`verify_host_key: :accept_local_tunnel_and_new` or `verify_host_key: :except_local_tunnel_and_new`

3. :null
`verify_host_key: :never`

4. :secure
`verify_host_key: :secure` or `verify_host_key: :always`


I'm open to other suggestions or keeping the `:null, :lenient, :strict, :secure`, i'd just like to take the opportunity to find the best names

```
   StrictHostKeyChecking
             If this flag is set to yes, ssh(1) will never automatically add host keys to the ~/.ssh/known_hosts file, and refuses to connect to hosts whose host key has changed.  This
             provides maximum protection against trojan horse attacks, though it can be annoying when the /etc/ssh/ssh_known_hosts file is poorly maintained or when connections to new
             hosts are frequently made.  This option forces the user to manually add all new hosts.

             If this flag is set to ``accept-new'' then ssh will automatically add new host keys to the user known hosts files, but will not permit connections to hosts with changed host
             keys.  If this flag is set to ``no'' or ``off'', ssh will automatically add new host keys to the user known hosts files and allow connections to hosts with changed hostkeys
             to proceed, subject to some restrictions.  If this flag is set to ask (the default), new host keys will be added to the user known host files only after the user has con-
             firmed that is what they really want to do, and ssh will refuse to connect to hosts whose host key has changed.  The host keys of known hosts will be verified automatically
             in all cases.
```

jaredbeck:
Hi Miklós, I think there's a lot of value in having the symbols (:strict, :secure) map directly to the classes (Strict, Secure). Perhaps we can preserve this mapping *and* improve the names by renaming the classes. 

Regarding the specific names, I'll have to give it some thought. Some are easy; renaming Null to Never seems like a definite improvement. I'm not sure about the other classes.

mfazekas:
@jaredbeck sure i'm totally fine with renaming classes to match names. If we can give more explicit/better names than `strict`, `secure`, `lenient` that would be nice.

jaredbeck:
> i'm totally fine with renaming classes to match names. If we can give more explicit/better names than strict, secure, lenient that would be nice.

Great! How about this:

| *Old Class* | *New Class* | *verify_host_key:* |
| --- | --- | --- |
| Null | Never | :never |
| Lenient | AcceptNewOrLocalTunnel | :accept_new_or_local_tunnel |
| Strict | AcceptNew | :accept_new |
| Secure | Known | :known |

# My thought process

## `Secure`

- Suggested names: Secure, Always
- Docs: 

> Does a strict host verification, looking the server up in the known host files to see if a key has already been seen for this server. If this server does not appear in any host file, an exception will be raised (HostKeyUnknown). This is in contrast to the "Strict" class, which will silently add the key to your known_hosts file. If the server does appear at least once, but the key given does not match any known for the server, an exception will be raised (HostKeyMismatch).Otherwise, this returns true.

- Key concept: Host must *already* be known
- My suggestion: `Known`

## `Strict`

- Suggested names: Strict, AcceptNew
- Docs:

> Does a strict host verification, looking the server up in the known host files to see if a key has already been seen for this server. If this server does not appear in any host file, this will silently add the server. If the server does appear at least once, but the key given does not match any known for the server, an exception will be raised (HostKeyMismatch).Otherwise, this returns true.

- Key concept: Accepts and remembers new hosts
- My suggestion: `AcceptNew`, because I really like the correlation between this and the `ssh_config` option `StrictHostKeyChecking accept-new`, as you pointed out, Miklós

## `Lenient`

- Suggested names: Lenient, AcceptLocalTunnelAndNew
- Docs:

> Basically the same as the Strict verifier, but does not try to actually verify a connection if the server is the localhost and the port is a nonstandard port number. Those two conditions will typically mean the connection is being tunnelled through a forwarded port, so the known-hosts file will not be helpful (in general).

- Key concept: Local tunnel
- My suggestion: AcceptNewOrLocalTunnel - because it contains the string "AcceptNew" and thus more closely matches the name of its parent class. I'd use the word "Or" because it seems clearer that "New" and "LocalTunnel" are two different things.

mfazekas:
@jaredbeck thanks looks great to me, i agree with the changes you've suggested.

The only minor conern is that `verify_host_key: :known` doesn't read as nice as for example `verify_host_key: :never`

jaredbeck:
> The only minor conern is that verify_host_key: :known doesn't read as nice as for example verify_host_key: :never

I agree. I'm OK with `Always`. I wish we could call it `Strict`, in keeping with SSH's `StrictHostKeyChecking`, but that's not an option.

mfazekas:
> I agree. I'm OK with Always. I wish we could call it Strict, in keeping with SSH's StrictHostKeyChecking, but that's not an option.

Yes sounds great, yes i agree `strict` is not an option as we've already used for different meaning.

jaredbeck:
Hi Miklós, ready for review. As before, travis passed and appveyor failed.

Issue #611, opened 2018-06-11
Changing ecb to ctr under jruby: padding error


film42:
### Expected behavior

The ssh connection should establish.

### Actual behavior

After running updating our app's bundle, and attempting to deploy via capistrano, we found the following error:

```
padding error, need 2061131439 block 16
```

### System configuration

- net-ssh: 5.0.1
- Ruby version: jruby 9.1.12.0, jruby 9.1.16.0 (Seems to be all jruby versions, though).

### Example App

You must a establish a connection to an ssh cipher `aes128-ctr` which will convert to open ssl cipher: `aes-128-ctr`. Previously, this would map to `aes-128-ecb`. Under jruby, the `ctr` cipher results in the error above: `padding error, need 2061131439 block 16`.

```
require "net/ssh"
Net::SSH.start("some-host", "some-user") {}
```

This was changed here: https://github.com/net-ssh/net-ssh/commit/73f204526a0d48df05c8236943f43778d542b242#diff-150dc6ef4d541dccc030bac074a3b70dL29

Merged here: https://github.com/net-ssh/net-ssh/pull/570

And it looks like jruby has been failing CI for a while now.

This might end up being a bug in jruby-openssl, but wanted to open an issue in case others are running into this problem.

mfazekas:
@film42 Yep that change was an optimization to use OpenSSL provided native cipher, which is probably not correctly implemented on jruby's openssl libraries. I guess we'd need to add a conditional that uses the ruby impl on jruby.

Unfortunately we don't have many resources to report/report JRuby related bugs, that's why JRuby is failing in our CI. PR's welcome!

film42:
@mfazekas Thanks for merging #612. I was just wondering when you were planning on cutting a new release for `net-ssh`. Thanks!

mfazekas:
5.0.2 is now released

PR #612, opened 2018-06-13
fix net/ssh 5+ for jruby

abrandoned:
while this doesn't cleanup the specs for jruby yet (a few issues are being fixed in jruby-openssl with `set_key` on pkey); it does allow net/ssh > 5 to work on Jruby as the ctr ciphers are not yet available in jruby-openssl

fixes issue #611


film42:
:+1: Nice solution. Really simple as well.

Issue #619, opened 2018-07-24
Net SSH fails with ECDSA keys without private key file


agrare:
### Expected behavior

With a private key (e.g. `.ssh/id_rsa`) and no public key (e.d. `.ssh/id_rsa.pub`) net-ssh typically derives the public key from the private key [ref](https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/authentication/key_manager.rb#L233).  This works great for RSA type ssh keys but not for ECDSA type ssh keys.

### Actual behavior

With a private key of type `OpenSSL::PKey::EC`, calling `.public_key` returns a `OpenSSL::PKey::EC::Point` object.  When this is being written out to the `Net::SSH::Buffer` `.to_blob` is called on the key [ref](https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/buffer.rb#L393).

The `.to_blob` method is defined for `OpenSSL::PKey::EC` [ref](https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/transport/openssl.rb#L184-L189) but not for `OpenSSL::PKey::EC::Point` which leads to the following error:
```
Traceback (most recent call last):
       16: from (irb):2
       15: from /var/lib/gems/2.5.0/gems/net-ssh-5.0.2/lib/net/ssh.rb:246:in `start'
       14: from /var/lib/gems/2.5.0/gems/net-ssh-5.0.2/lib/net/ssh/authentication/session.rb:71:in `authenticate'
       13: from /var/lib/gems/2.5.0/gems/net-ssh-5.0.2/lib/net/ssh/authentication/session.rb:71:in `each'
       12: from /var/lib/gems/2.5.0/gems/net-ssh-5.0.2/lib/net/ssh/authentication/session.rb:85:in `block in authenticate'
       11: from /var/lib/gems/2.5.0/gems/net-ssh-5.0.2/lib/net/ssh/authentication/methods/publickey.rb:19:in `authenticate'
       10: from /var/lib/gems/2.5.0/gems/net-ssh-5.0.2/lib/net/ssh/authentication/key_manager.rb:119:in `each_identity'
        9: from /var/lib/gems/2.5.0/gems/net-ssh-5.0.2/lib/net/ssh/authentication/key_manager.rb:119:in `each'
        8: from /var/lib/gems/2.5.0/gems/net-ssh-5.0.2/lib/net/ssh/authentication/key_manager.rb:122:in `block in each_identity'
[...]

PR #620, opened 2018-07-24
Implement to_blob on OpenSSL::PKey::EC::Point

agrare:
Fix an issue where writing an ECDSA public_key out to a Net::SSH::Buffer
fails when calling to_blob on the key due to the method being undefined.

Fixes https://github.com/net-ssh/net-ssh/issues/619


agrare:
:tada: thanks @mfazekas !

PR #637, opened 2018-10-13
Prefer more modern cryptographic algorithms

### LINKED ISSUE #627: Prefer modern cryptographic algorithms by default
The current algorithm preference order is inconsistent with the defaults from OpenSSH. For example, https://github.com/net-ssh/net-ssh/blob/36c0b46eeb3c3831ce805566cbb0ef26dfdca613/lib/net/ssh/transport/algorithms.rb#L28 prefers old host key algorithms ssh-rsa and even ssh-dss over ecdsa-sha2-nistp256, which is OpenSSH's current preferred algorithm. Obviously, an exact match with OpenSSH's defaults now and in the future is infeasible, but there are some clear security wins that can be made simply by reordering this list.

Are these defaults causing user-visible issues? Yes, that's why I'm here. Chef, and therefore Knife, does not allow overrides of the Net::SSH defaults here. This causes key mismatch errors when "knife ssh" creates a host key entry for an RSA key and a user or script on the same machine uses an OpenSSH client with default settings, which negotiates a modern host key algorithm. My workaround to this has been to specify OpenSSH's default host key ordering in my SSH config, which is awkward.

Even without user-visible issues, I would argue that Net::SSH should prefer stronger algorithms by default to protect users of the library that don't know which algorithms they should be using.

torrancew:
This commit modifies Net::SSH to prefer strong encryption for HMAC,
Cipher, Host Key Authentication and Key Exchange operations. It should resolve #627.

### DIFF
--- a/lib/net/ssh/transport/algorithms.rb
+++ b/lib/net/ssh/transport/algorithms.rb
@@ -26,35 +26,44 @@ module Net
         # Define the default algorithms, in order of preference, supported by
         # Net::SSH.
         ALGORITHMS = {
-          host_key: %w[ssh-rsa ssh-dss
-                       ssh-rsa-cert-v01@openssh.com
-                       ssh-rsa-cert-v00@openssh.com],
-          kex: %w[diffie-hellman-group-exchange-sha1
-                  diffie-hellman-group1-sha1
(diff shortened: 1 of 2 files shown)


mfazekas:
@torrancew thanks for the PR

PR #641, opened 2018-11-08
Server key signature verification done even if not requested

pkliczewski:
In my environment I know that server key verification would fail so
I set `verify_host_key` to `:never` I see following error:

```
diffie_hellman_group1_sha1.rb:211:in `verify_signature': could not verify server signature (Net::SSH::Exception)
```

This PR updates the Verifiers to provide key verification as well as key signature so when setting
that this operation should never be run it works fine.

### DIFF
--- a/lib/net/ssh/transport/session.rb
+++ b/lib/net/ssh/transport/session.rb
@@ -274,6 +274,23 @@ module Net
 
         private
 
+        # Compatibility verifier which allows users to keep using
+        # custom verifier code without adding new :verify_signature
+        # method.
+        class CompatibleVerifier
+          def initialize(verifier)
+            @verifier
+          end
+
+          def verify(arguments)
+            @verifier.verify(arguments)
+          end
+
+          def verify_signature(&block)
+            yield
+          end
+        end
+
         # Instantiates a new host-key verification class, based on the value of
         # the parameter.
         #
@@ -285,8 +302,8 @@ module Net
         # - :accept_new (insecure)
         # - :always (secure)
         #
-        # If the argument happens to respond to :verify, it is returned
-        # directly. Otherwise, an exception is raised.
+        # If the argument happens to respond to :verify and :verify_signature,
+        # it is returned directly. Otherwise, an exception is raised.
         #
         # Values false, true, and :very were deprecated in
         # [#595](https://github.com/net-ssh/net-ssh/pull/595)
@@ -314,7 +331,12 @@ module Net
             Net::SSH::Verifiers::Always.new
           else
             if verifier.respond_to?(:verify)
-              verifier
+              if verifier.respond_to?(:verify_signature)
+                verifier
+              else
+                Kernel.warn("Warning: verifier without :verify_signature is deprecated")
+                CompatibleVerifier.new(verifier)
+              end
             else
               raise(
                 ArgumentError,
--- a/lib/net/ssh/verifiers/accept_new.rb
+++ b/lib/net/ssh/verifiers/accept_new.rb
@@ -21,6 +21,13 @@ module Net
             return true
           end
         end
+
+        def verify_signature(&block)
+          yield
+        rescue HostKeyUnknown => err
+          err.remember_host!
+          return true
+        end
       end
 
     end
--- a/lib/net/ssh/verifiers/always.rb
+++ b/lib/net/ssh/verifiers/always.rb
@@ -34,6 +34,10 @@ module Net
           found
         end
 
+        def verify_signature(&block)
+          yield
+        end
+
         private
 
         def process_cache_miss(host_keys, args, exc_class, message)
--- a/lib/net/ssh/transport/kex/diffie_hellman_group1_sha1.rb
+++ b/lib/net/ssh/transport/kex/diffie_hellman_group1_sha1.rb
@@ -206,7 +206,7 @@ module Net
 
             hash = @digester.digest(response.to_s)
 
-            raise Net::SSH::Exception, "could not verify server signature" unless result[:server_key].ssh_do_verify(result[:server_sig], hash)
+            raise Net::SSH::Exception, "could not verify server signature" unless connection.host_key_verifier.verify_signature { result[:server_key].ssh_do_verify(result[:server_sig], hash) }
 
             return hash
           end
--- a/lib/net/ssh/verifiers/never.rb
+++ b/lib/net/ssh/verifiers/never.rb
@@ -10,6 +10,10 @@ module Net
         def verify(arguments)
           true
         end
+
+        def verify_signature(&block)
+          true
+        end
       end
 
     end
--- a/test/common.rb
+++ b/test/common.rb
@@ -74,6 +74,10 @@ class MockTransport < Net::SSH::Transport::Session
     def verify(data)
       @block.call(data)
     end
+
+    def verify_signature(&block)
+      yield
+    end
   end
 
   attr_reader :host_key_verifier
--- a/test/transport/test_session.rb
+++ b/test/transport/test_session.rb
@@ -92,8 +92,15 @@ module Transport
     end
 
     def test_verify_host_key_value_responding_to_verify_should_pass_muster
-      object = stub("thingy", verify: true)
+      object = stub("thingy", verify: true, verify_signature: true)
       assert_equal object, session(verify_host_key: object).host_key_verifier
+    end
+
+    def test_deprecated_host_key_verifier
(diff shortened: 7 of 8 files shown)


pkliczewski:
@mfazekas please take a look whether it makes sense.

mfazekas:
@pkliczewski another issue is that we right now allow any object to be a verifier if it repsponds to `verify`. We'd need to document new method and for backward compat we'd need to have a default `verify_signature` impl for custom verifiers.

https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/transport/session.rb#L316-L323

pkliczewski:
@mfazekas Here is how I found this issue and how I checked it. I am working on sync folders action for [vagrant-kubevirt](http://github.com/pkliczewski/vagrant-kubevirt/) provider. I implemented `vagrant ssh` already which works nicely. [Here](https://github.com/hashicorp/vagrant/blob/85acf0cac724ef4bfda593a66e0c17f7e4110da0/lib/vagrant/util/ssh.rb) is the code which is not using `net-ssh`. I based my sync folders implementation on vagrant synced_folders action which is using [this](https://github.com/hashicorp/vagrant/blob/7efd501fe80c57d8a9138342a6545e0cfaf032be/plugins/communicators/ssh/communicator.rb) code. I can confirm that it works using plain `ssh` and `vagrant ssh` where as sync folder fails with signature verification issue even though I set `verify_host_key` to `:never`.

pkliczewski:
@mfazekas Backward compatibility should be ok due to [this change](https://github.com/net-ssh/net-ssh/pull/641/files#diff-8d3660909f2b668f151a8fc67bfebfaeR37). All Accept* verifiers inherit from `Always` which provides default implementation. Where should I document new method?

mfazekas:
> @mfazekas Backward compatibility should be ok due to [this change](https://github.com/net-ssh/net-ssh/pull/641/files#diff-8d3660909f2b668f151a8fc67bfebfaeR37). All Accept* verifiers inherit from `Always` which provides default implementation. Where should I document new method?

In previous version users can provide it's own implementation provided it implements `verify`. See https://github.com/net-ssh/net-ssh/blob/dd13dd44d68b7fa82d4ca9a3bbe18e30c855f1d2/lib/net/ssh/transport/session.rb#L316-L318

pkliczewski:
@mfazekas I updated `Session#select_host_key_verifier` doc with new requirement. Is this enough or is there dev documentation that I should update?

pkliczewski:
@mfazekas Please review and let me know what is needed to finalize this PR

lib/net/ssh/transport/session.rb
```diff
@@ -313,7 +313,7 @@ def select_host_key_verifier(verifier)
           when :always then
             Net::SSH::Verifiers::Always.new
           else
-            if verifier.respond_to?(:verify)
+            if verifier.respond_to?(:verify) and verifier.respond_to?(:verify_signature)
```
mfazekas: please use `&&` also this means it's a backward incompatible change so we need to release as 6.0

lib/net/ssh/verifiers/accept_new.rb
```diff
@@ -21,6 +21,15 @@ def verify(arguments)
             return true
           end
         end
+
+        def verify_signature(arguments)
+          begin
```
mfazekas: Should work without begin/end

lib/net/ssh/verifiers/accept_new_or_local_tunnel.rb
```diff
@@ -17,6 +17,10 @@ def verify(arguments)
           super
         end
 
+        def verify_signature(arguments)
```
mfazekas: I guess we can remove this method

lib/net/ssh/transport/session.rb
pkliczewski: ok sounds like breaking change. Is there any specific time frame when this release will be available?

pkliczewski:
@mfazekas Your comments fixed. Please review

pkliczewski:
@mfazekas anything else you want me to fix for this PR?

lib/net/ssh/verifiers/accept_new.rb
pkliczewski: Done

lib/net/ssh/verifiers/accept_new_or_local_tunnel.rb
pkliczewski: Done

lib/net/ssh/transport/session.rb
mfazekas: What  about doing soemthing like this:

```ruby
class CompatibleVerifier
  def initialize(verifier)
     @verifier
  end
  def verify_signature(key: , signature; ,hash: )
     return key.ssh_do_verify(signature, hash)
  end
end
...
if verifier.respond_to?(:verify)
  if verifyer.response_to?(:verify_signature)
    verifyer  
  else
     log "Warning: verifier without verify_signature is deprecated"
     CompatibleVerifier.new(verifier)
  end
end

```

lib/net/ssh/verifiers/always.rb
```diff
@@ -34,6 +34,10 @@ def verify(arguments)
           found
         end
 
+        def verify_signature(arguments)
+          return arguments[:key].ssh_do_verify(arguments[:signature], arguments[:hash])
```
mfazekas: I think we can use keyword arguments as net-ssh is ruby2.2+ i think

lib/net/ssh/transport/session.rb
pkliczewski: Make sense, will update.

lib/net/ssh/verifiers/always.rb
pkliczewski: OK, will update

mfazekas:
@pkliczewski I'm also thinking about veriy_signature method, if it needs to know about key, signature,hash. 

I'm thinking something like along the lines:


```
def verify_signature(&block)
   yield
end

def verify_signature(&block)
   yield
end

def verify_signature(&block)
   true
end

def verify_signature(&block)
    yield
rescue HostKeyUnknown => err
     err.remember_host!
     return true
 end

Usage would be:

raise Net::SSH::Exception, "could not verify server signature" 
  unless verifier.verify_signature { result[:server_key].ssh_do_verify(result[:server_sig], hash) }
```

What do you think?
For custom verifier it would simplify the docs, etc. But makes a code a bit more complicated.

pkliczewski:
Sounds good, let's do it!

lib/net/ssh/transport/session.rb
pkliczewski: I pushed the update with all the changes you asked. Please take a look.

lib/net/ssh/transport/session.rb
```diff
@@ -274,6 +274,16 @@ def hint(which, value=true)
 
         private
 
+        class CompatibleVerifier
+          def initialize(verifier)
+            @verifier
+          end
+
```
mfazekas:
```suggestion

def verify(arguments)
   @verifier.verify(arguments)
end
```

lib/net/ssh/transport/session.rb
```diff
@@ -314,7 +324,12 @@ def select_host_key_verifier(verifier)
             Net::SSH::Verifiers::Always.new
           else
             if verifier.respond_to?(:verify)
-              verifier
+              if verifier.respond_to?(:verify_signature)
+                verifier
+              else
+                log "Warning: verifier without verify_signature is deprecated"
```
mfazekas:
```suggestion
                Kernel.warn("Warning: verifier without :verify_signature is deprecated")
```

lib/net/ssh/transport/session.rb
pkliczewski: Applied

lib/net/ssh/transport/session.rb
pkliczewski: Applied

pkliczewski:
@mfazekas Now codecov-io is happy. Is there anything I need to improve?

mfazekas:
@pkliczewski thanks much for the PR!

pkliczewski:
:tada: Thanks, When this can be released?

gytisgreitai:
Hi, is the usage of `verify_signature` documented somewhere? The http://net-ssh.github.io/net-ssh/Net/SSH.html#method-c-start does seem to state that `You can also provide an own Object which responds to verify. The argument given to verify` yet nothing about `verify_signature` but nothing on the signature and how to use that?

PR #642, opened 2018-11-09
Improve processing of ssh_config Match conditions

### LINKED ISSUE #631: NetSSH matches all conditions if ssh config has `Match` with criteria other than `host`
### System configuration

- net-ssh version: `5.0.2`
- Ruby version: `ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-darwin17]`
- `ssh -v`: `OpenSSH_7.6p1, LibreSSL 2.6.2`

### Example App

I am using [`Match`](https://man.openbsd.org/ssh_config#Match) in my ssh config file with multiple blocks with criteria `exec`. 

From the case statement [here](https://github.com/net-ssh/net-ssh/blob/7429a290230b2f31179047a7130b40744e1b0b54/lib/net/ssh/config.rb#L335), looks like net-ssh does not support criteria other than `host`. The issue is that the even for unsupported blocks, because of the default=true [here](https://github.com/net-ssh/net-ssh/blob/7429a290230b2f31179047a7130b40744e1b0b54/lib/net/ssh/config.rb#L333), NetSSH sources blocks for each of these criteria and we end up with mixed up config for connection.

I feel that for unsupported blocks / criteria, we should just ignore the block ( as if they are absent ) (ie have `eval_match_conditions` return false for unsupported blocks)

Sample ssh config:
```bash
IdentityFile ~/.ssh/id_rsa

Host random_host
  Port 4444

Match exec "exit 1"
  ProxyJump other_host

Match exec "exit 1"
  IdentityFile ~/.ssh/personal_id_rsa
```

Sample net/ssh code:
```ruby
require 'net/ssh'

Net::SSH.start("random_host", "dev", :verbose=>:debug) do |ssh|
  ssh.exec!("ls /tmp")
  ssh.loop
end
```

### Expected behavior

Should not use proxy for connection to random:host:4444 with `id_rsa`
- cli ssh evaluates exec conditions are executes based on return code
- This issue is not for implementing `exec` support, but just to have false as default so as to ignore unsupported blocks

### Actual behavior

Output from the above code
```bash
DEBUG -- net.ssh.transport.session[3fc610c9c76c]: establishing connection to xx.xx.xx.xx:4444 through proxy
```
Also ^ `personal_id_rsa` is used instead of `id_rsa`

aleksandrs-ledovskis:
Previously wasn't correctly handling quoted and `=` delimited values.
Also, fixes logic that unsupported `Match` conditions result in subsequent declarations to be applied for configuration.

OpenSSH implementation details:
- https://github.com/openssh/openssh-portable/blob/624d19ac2d56fa86a22417c35536caceb3be346f/readconf.c#L599
- https://github.com/openssh/openssh-portable/blob/624d19ac2d56fa86a22417c35536caceb3be346f/misc.c#L284-L285

---

Closes #631

### DIFF
--- a/lib/net/ssh/config.rb
+++ b/lib/net/ssh/config.rb
@@ -96,7 +96,7 @@ module Net
             next if value.nil?
 
             key.downcase!
-            value = $1 if value =~ /^"(.*)"$/
+            value = unquote(value)
 
             value = case value.strip
                     when /^\d+$/ then value.to_i
@@ -328,12 +328,17 @@ module Net
         end
 
         def eval_match_conditions(condition, host, settings)
-          conditions = condition.split(/\s+/)
+          # Not using `\s` for whitespace matching as canonical
+          # ssh_config parser implementation (OpenSSH) has specific character set.
+          # Ref: https://github.com/openssh/openssh-portable/blob/2581333d564d8697837729b3d07d45738eaf5a54/misc.c#L237-L239
+          conditions = condition.split(/[ \t\r\n]+|(?<!=)=(?!=)/).reject(&:empty?)
           return true if conditions == ["all"]
 
           conditions = conditions.each_slice(2)
-          matching = true
+          condition_matches = []
           conditions.each do |(kind,exprs)|
+            exprs = unquote(exprs)
+
             case kind.downcase
             when "all"
               raise "all cannot be mixed with other conditions"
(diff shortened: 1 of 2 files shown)


lib/net/ssh/config.rb
```diff
@@ -346,13 +351,17 @@ def eval_match_conditions(condition, host, settings)
               exprs.split(",").each do |expr|
                 condition_met = condition_met || host =~ pattern2regex(expr)
               end
-              matching = matching && negated ^ condition_met
+              matching = true && negated ^ condition_met
```
mfazekas: I don't think it's correct, when there are multiple condiitons, for example:

```ruby
Match host *.home.lan host !foo.home.lan
```

I think your code will condsider google.com as match.

mfazekas:
@aleksandrs-ledovskis Thanks for the PR looks good.

lib/net/ssh/config.rb
aleksandrs-ledovskis: Indeed, that was missed. Have updated logic to correctly handle this situation and added relevant test.

aleksandrs-ledovskis:
Travis build is borked (seems due to not-yet-merged changes in #646)

aleksandrs-ledovskis:
Build is green. Seems LGTM

PR #644, opened 2018-11-09
fix deprecated use of verify_host_key

klausbadelt:
Fix `verify_host_key: false is deprecated, use :never` when using `Net::SSH::Test` in your tests.

Avoids lots of these warnings, for example, when running `net-ssh/net-sftp` tests, as we had in net-ssh/net-sftp#82.


mfazekas:
@klausbadelt thanks for the PR, i think this requires Net-ssh 5.0 or later, can you modify the dependencies also?

mfazekas:
@klausbadelt sorry for my last comment i've now realized this is net-ssh not net-sftp.

PR #645, opened 2018-11-25
Add support for IdentityAgent in ssh_config

frenkel:
This commit adds support for configuring the IdentityAgent through an SSH config file.

### DIFF
--- a/lib/net/ssh/authentication/agent.rb
+++ b/lib/net/ssh/authentication/agent.rb
@@ -62,9 +62,9 @@ module Net
 
         # Instantiates a new agent object, connects to a running SSH agent,
         # negotiates the agent protocol version, and returns the agent object.
-        def self.connect(logger=nil, agent_socket_factory = nil)
+        def self.connect(logger=nil, agent_socket_factory = nil, identity_agent = nil)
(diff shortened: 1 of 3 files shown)


mfazekas:
@frenkel  thanks for the PR!

vrinek:
@frenkel @mfazekas I think this PR may have missed one point: `Net::SSH::VALID_OPTIONS`.

I get an `ArgumentError: invalid option(s): identity_agent` exception in my local setup when I have an `IdentityAgent` option in my `~/.ssh/config`.

Would you say this is relevant or should I file it as a separate issue?

frenkel:
I'm not sure, I don't have this error while I use it on a number of computers. Is everything up to date?

fwininger:
@vrinek can you give us your ruby version and the version of net/ssh you use ?

and if you have the complete stack error, this would be awesome.

vrinek:
I'm using `net-ssh` via `capistrano` which gives me `ArgumentError: invalid option(s): identity_agent` for each remote server it's trying to SSH to.

Versions:
Ruby: 2.4.7
Net/SSH: 5.2.0

Issue #638, opened 2018-10-25
Argument error: expected 64-byte String, got 3


ekampp:
### Expected behavior

I'm able to connect to the server

I have recently begun upgrading my ssh keys to the new ECDSA, and I got a helpful message about adding the `ed25519` and `bcrypt_pbkdf` gems which I did. After this I reached this error, which don't seem to be super helpful.

### Actual behavior

Generating a key with `ssh-keygen -b 4096 -C "<USER>@<IP>"`

I get an `ArgumentError` with: "expected 64-byte String, got 3"

If I generate the key specifically with `-t ecdsa -b 512`

I get "expected 64-byte String, got 133"

### System configuration

- net-ssh version: 5.0.2
- Ruby version: 2.5.1p57

### Example App

```
require 'net/ssh'

opts = {
  :host => '<IP>',
  :port => 22,
  :user => '<USER>'
}

opts[:logger] = Logger.new(STDERR)
opts[:logger].level = Logger::INFO

Net::SSH.start(opts[:host], opts[:user], :keys_only => true, :logger => opts[:logger]) do |ssh|
  ls_result = ssh.exec("ls -l")
  puts ls_result
end
```

stiller-leser:
Not a solution to this issue, but a workaround from https://github.com/duplicati/duplicati/issues/3360#issuecomment-418826330. Unfortunately this requires you to generate a new key in the old key format using ` ssh-keygen -m PEM -t rsa -b 4096`. The reasoning for the switch to the new format can be found here: https://https.www.google.com.tedunangst.com/flak/post/new-openssh-key-format-and-bcrypt-pbkdf

SharkFourSix:
> Not a solution to this issue, but a workaround from [duplicati/duplicati#3360 (comment)](https://github.com/duplicati/duplicati/issues/3360#issuecomment-418826330). Unfortunately this requires you to generate a new key in the old key format using ` ssh-keygen -m PEM -t rsa -b 4096`. The differences are described here: https://https.www.google.com.tedunangst.com/flak/post/new-openssh-key-format-and-bcrypt-pbkdf
> 
> There is also a discussion about it here #633, with the bug apparently located here: https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/key_factory.rb#L112

Sadly this does not seem to work.

mfazekas:
@SharkFourSix this should be fixed in 5.2.0, isn't it?

quickdudley:
[...]

Issue #633, opened 2018-09-23
Cannot ssh with ssh RSA keys having BEGIN OPENSSH PRIVATE KEY header (PKCS8 format)


frezbo:
### Expected behavior

SSH should work

### Actual behavior

SSH authentication fails, but manual ssh works

### System configuration

- net-ssh version - 5.0.2
- Ruby version - ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-linux]

### Example App

```
net_ssh_test (aws:rean-devops)(kc)$ bundle list 
Gems included by the bundle:
  * bcrypt_pbkdf (1.0.0)
  * bundler (1.16.5)
  * ed25519 (1.2.4)
  * net-ssh (5.0.2)
net_ssh_test (aws:rean-devops)(kc)$ 
```

key generated on Fedora 28 with `ssh-keygen -q -N '' -f image-keypair`

System details:
```
net_ssh_test (aws:rean-devops)(kc)$ uname -a
Linux frezbo 4.18.8-200.fc28.x86_64 #1 SMP Sun Sep 16 18:15:41 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
net_ssh_test (aws:rean-devops)(kc)$ cat /etc/*-release
Fedora release 28 (Twenty Eight)
NAME=Fedora
VERSION="28 (Workstation Edition)"
ID=fedora
VERSION_ID=28
PLATFORM_ID="platform:f28"
PRETTY_NAME="Fedora 28 (Workstation Edition)"
ANSI_COLOR="0;34"
CPE_NAME="cpe:/o:fedoraproject:fedora:28"
HOME_URL="https://fedoraproject.org/"
SUPPORT_URL="https://fedoraproject.org/wiki/Communicating_and_getting_help"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=28
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=28
PRIVACY_POLICY_URL="https://fedoraproject.org/wiki/Legal:PrivacyPolicy"
VARIANT="Workstation Edition"
VARIANT_ID=workstation
Fedora release 28 (Twenty Eight)
Fedora release 28 (Twenty Eight)
net_ssh_test (aws:rean-devops)(kc)$ 
```

[...]

PR #646, opened 2018-11-27
Support new private key format for other than ed25519 keys

mfazekas:
Fixes: #638 
Fixes: #633


akostadinov:
I have a RSA key in the new format. I've installer net-ssh from git by bundle. Now the key is read. But still connection to machine fails. While `ssh -i ... machine` on command line works.

mfazekas:
@akostadinov can you test again? It should be fixed for rsa keys?

akostadinov:
Can try Monday, thank you!

PR #647, opened 2018-12-04
Clear SSH_AUTH_SOCK for duration of test run

aleksandrs-ledovskis:
Permits per-scenario/test setup in usual `setup/teardown` constructs
Issue was spotted on local development machine where `SSH_AUTH_SOCK` was
present in ENV list - it messed up with "test/authentication/test_session.rb" tests


aleksandrs-ledovskis:
Travis build seems paused, but that is likely due to troubles at their side

image from https://www.traviscistatus.com/

Ref: https://twitter.com/robbieaverill/status/1069929889107726336

PR #656, opened 2019-01-29
Fix host key checking

smortex:
The known_hosts file may contain keys associated with a hostname, an
ip-address, or both.

When validating a key, the net-ssh gem ensure that both the hostname and
the ip-address match beforce adding a key.  Thus, if the known_hosts
file only contains one of these two pieces of information, the host key
verification fails.

Instead of adding keys when both the hostname and the ip-address match,
add them when any of these information match.


smortex:
Hi @mfazekas!  My previous attempt to add tests failed because of another bug.  A fresh view allowed me to spot (b5af8a8) it and add a few tests.

It also allowed me to see a difference between net-ssh and ssh(1), I detailed the issue in 855d617.

Globing is also still not handled for matching i=entries in the known_hosts file, but I think this can be part of another PR (bugfix ≠ adding missing features).

lib/net/ssh/transport/session.rb
```diff
@@ -105,16 +105,6 @@ def host_as_string
           @host_as_string ||= begin
             string = "#{host}"
             string = "[#{string}]:#{port}" if port != DEFAULT_PORT
-
-            peer_ip = socket.peer_ip
-
-            if peer_ip != Net::SSH::Transport::PacketStream::PROXY_COMMAND_HOST_IP &&
```
mfazekas: Are you sure we don't need this?
Looking at OpenSSH code, it seems to be resolve the address then considers it's a match if either ip or hostname matches.

https://github.com/openssh/openssh-portable/blob/master/sshconnect.c#L665-L672

https://github.com/openssh/openssh-portable/blob/c327813ea1d740e3e367109c17873815aba1328e/clientloop.c#L2133-L2144

https://github.com/openssh/openssh-portable/blob/2f131e1b34502aa19f345e89cabf6fa3fc097f09/hostfile.c#L723-L749

smortex: I have to admin that I have not compared the code with openssh's, only seen that host I could connect to by `ssh host` where not accessible using the gem.

Adding `,1.2.3.4` after the hostname in the known_hosts files unbreaked the gem.  So my first attempt was to make the gem work in both cases (`ssh host` and `ssh 1.2.3.4`) when the known_host entry was not in the form `hostname,ip`.

Then I saw that `ssh host2` (with host2 resolving to 1.2.3.4) was not working with plain ssh.  So I felt it was better to have the same behavior, that is not resolve the host ip address and attempt to find an entry using this ip address.  Either the user want to connect to "host" and a "host" key is looked for, either she wants to connect to "1.2.3.4" and a key for "1.2.3.4" is looked for.

I am open to some adjustment of course.

All this was tested with OpenSSH 7.5, not the latest OpenSSH that may behave differently.

donoghuc:
This seems like a really nice improvement. Would love to see it adopted.

lib/net/ssh/transport/session.rb
mfazekas: @smortex  thanks much for the effort. Sorry for the slow reactions but i need to make sure we don't break anything with the changes.

I've played with openssh a bit (using `OpenSSH_7.8p1, LibreSSL 2.7.3`) these are my observations. (I've played with heroku.com and `50.19.85.154 heroku-ssh1.dev` and `50.19.85.154 heroku-ssh2.dev` in /etc/hosts).

- ssh adds both name and resolved ip to known_hosts file (`ssh heroku.com: heroku.com,50.19.85.154 ssh-rsa AA `)
- ssh does recognize the host by name or ip  (so both `ssh heroku.com` and `ssh 50.19.85.154`  works)
- ssh by different name but same ip does not work (so `ssh heroku-ssh1.dev` is a new entry, `heroku-ssh1.dev ssh-rsa AAAAB`). Note that the new entry added without ip this time.

Also if i clear /etc/hosts file and try again
- ssh adds both name and resolved ip to known_hosts file (`ssh heroku-ssh1.dev:  heroku-ssh1.dev,50.19.85.154 ssh-rsa AA `)
- if i change dns to 50.19.85.156 in `/etc/hosts` then it shows a warning and adds a new entry with just the ip address (`ssh heroku-ssh1.dev: 50.19.85.156 ssh-rsa AA` and `Warning: Permanently added the RSA host key for IP address '50.19.85.156' to the list of known hosts.`)
- if i change the entry in `/etc/hosts/` for 50.19.85.156 then `ssh heroku-ssh1.dev` shows a warning:  `
Warning: the RSA host key for 'heroku-ssh1.dev' differs from the key for the IP address '50.19.85.156'
Offending key for IP in /Users/boga/.ssh/known_hosts:7
Matching host key in /Users/boga/.ssh/known_hosts:5
`

This is probably the `CheckHostIP` option:
```
    CheckHostIP
             If set to yes (the default), ssh(1) will additionally check the host IP address in the known_hosts file.  This allows it to detect if a host key changed due to DNS spoof-
             ing and will add addresses of destination hosts to ~/.ssh/known_hosts in the process, regardless of the setting of StrictHostKeyChecking.  If the option is set to no, the
             check will not be executed.
```

smortex: Ah!  The default has changed :wink: Here is what `ssh_config(5)` says on FreeBSD 11.2:

```
     CheckHostIP                                                                
             If set to yes, ssh(1) will additionally check the host IP address
             in the known_hosts file.  This allows it to detect if a host key
             changed due to DNS spoofing and will add addresses of destination
             hosts to ~/.ssh/known_hosts in the process, regardless of the
             setting of StrictHostKeyChecking.  If the option is set to no,
             the check will not be executed.  The default is no.
```

This probably explains why I could not manage to use bolt which relies on `net-ssh` without tweaking it.  My known_hosts file does not have _any_ `,1.2.3.4`.

I guess that when matching a line, we can see if it has the form:

* `hostname`: ensure it matches and continue;
* `hostname:ip`: resolve hostname, and ensure both hostname and ip match.

I don't know what to think about checking if `CheckHostIP` is set in the configuration because if absent behavior should depend on the version of OpenSSH installed; and behavior of the same program on different systems would be different, which is probably a terrible idea.  Any though?

mfazekas: I'm not sure about usefullness of `CheckHostIP`, it sounds not too usefull for me, in today's cloud world where ip's are rarely fixed. But this is what current net-ssh implements - mostly.
Yes if the known_hosts has no hostname then we can accept a hostname only match, or we can at least show a nicer error/warning telling that the host was found but with different ip.
We can also add a new check_host_ip option and when it's true we reject if the ip doesn't match and when it's false we don't care about ip. 

Unfortunately we do allow custom known host implementations - see https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/transport/session.rb#L97-L98 and https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh.rb#L121-L123 so we cannot change `Net::SSH::Transport#host_as_string` as that would require a major version bump beacause of semver.

smortex: I see.  I'll remove bc86a70b9c7bce80a07d53109e4bd973d890a0fe and add support for `check_host_ip` defaulting to `false` as you suggest.  Probably next week.

Thanks!

smortex: Done!

Because of reverting some commits, I squashed the previous work with the new changes.  This requires a review from scratch, but since a lot of clutter got trimmed, I believe this is less work :smile: 

Resolving this conversation, please comment in the new commit diff.

Thanks!

mfazekas:
@smortex thanks much for the PR, i'll probably change check_host_ip default to true in 5.1.1, and will change to false in next major version.

donoghuc:
@mfazekas When do you expect to release 5.1.1? Would like to use this! Thanks.

PR #669, opened 2019-03-22
Pass :password_prompt to KeyFactory.load_private_key

### LINKED ISSUE #668: The password_prompt option is ignored
### Expected behavior

`Net::SSH.start(host, user, password_prompt: CustomPromptClass.new)` would be expected to utilize the custom prompter when a passphrase prompt is displayed.

### Actual behavior

The default `Net::SSH::Prompt` / `Net::SSH::Prompt::Prompter` is used instead.

### System configuration

- Ruby 2.6.1
- Net::SSH 5.2.0
- (using ED25519 key)

### Example App

```ruby
gem 'net-ssh'
require 'net/ssh'

@host = 'localhost'
@user = ENV['USER']

class CustomPrompt
  class Prompter
    def ask(*opts)
      puts "just asking.."
    end
  end

  def start(*opts)
    Prompter.new
  end
end

Net::SSH.start(@host, @user, password_prompt: CustomPrompt.new) do |ssh|
end
```

The regular `Net::SSH::Prompt::Prompter` is used instead of the custom one.

kke:
Fixes #668

At least in the issue example app's simple case.

The example app from issue

#### Before:

```
$ ruby example.rb
Enter passphrase for /Users/kimmo/.ssh/passphrase:
```

#### After:

```
$ ruby example.rb
just asking..
just asking..
just asking..
```

### DIFF
--- a/lib/net/ssh/authentication/key_manager.rb
+++ b/lib/net/ssh/authentication/key_manager.rb
@@ -139,7 +139,7 @@ module Net
 
           if info[:key].nil? && info[:from] == :file
             begin
-              info[:key] = KeyFactory.load_private_key(info[:file], options[:passphrase], !options[:non_interactive])
+              info[:key] = KeyFactory.load_private_key(info[:file], options[:passphrase], !options[:non_interactive], options[:password_prompt])
             rescue OpenSSL::OpenSSLError, Exception => e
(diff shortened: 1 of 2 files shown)


mfazekas:
@kke thanks looks great! Would it be hard to write a testcase for it as well?

kke:
Test added, fails on master, passes on PR.

mfazekas:
@kke thanks for the PR! sorry for the long turnaround

PR #671, opened 2019-03-23
Deprecating RC4 as per https://tools.ietf.org/html/draft-ietf-curdle-rc4-die-die-die-10

jmutkawoa:


loganaden:
Any update on this PR ? RC4 is being phased out on the net.

mfazekas:
@loganaden Thanks  - this will be in next major release 6.0

jmutkawoa:
@mfazekas thanks for taking into consideration

PR #672, opened 2019-04-26
Send EOF on closing channels.

misdoro:
Sending EOF is the default behaviour of openssh client.
Some servers (SSH-2.0-7.9.0.0_openssh GlobalSCAPE) ignore the CLOSE message, but respond to EOF by closing the channel.
OpenSSH server handles the EOF properly.

Might fix the https://github.com/net-ssh/net-sftp/issues/22 and similar hangs on closing the SFTP connection.

### DIFF
--- a/lib/net/ssh/connection/session.rb
+++ b/lib/net/ssh/connection/session.rb
@@ -120,7 +120,12 @@ module Net
         # connection.
         def close
           info { "closing remaining channels (#{channels.length} open)" }
-          channels.each { |id, channel| channel.close }
+
+          channels.each do |id, channel|
+            channel.eof!
+            channel.close
+          end
+
           begin
             loop(0.1) { channels.any? }
           rescue Net::SSH::Disconnect


mfazekas:
@misdoro thanks for the PR, and sorry for the late reply

Reading at SSH rfc:
https://tools.ietf.org/html/rfc4254

```
When either party wishes to terminate the channel, it sends
   SSH_MSG_CHANNEL_CLOSE.  Upon receiving this message, a party MUST
   send back an SSH_MSG_CHANNEL_CLOSE unless it has already sent this
   message for the channel.  The channel is considered closed for a
   party when it has both sent and received SSH_MSG_CHANNEL_CLOSE, and
   the party may then reuse the channel number.  A party MAY send
   SSH_MSG_CHANNEL_CLOSE without having sent or received
   SSH_MSG_CHANNEL_EOF.
```
So according to protocol we don't have to send eof before close.

The changes looks innocent, but i'd like to see some proof (logs?) that it solves the issue with GlobalSpace server and that openssh sends eof before close during normal sessions.

misdoro:
Let's try to resuscitate it.

Looking at https://github.com/openssh/openssh-portable/blob/master/nchan.c#L56 sending EOF was mandatory in a previous version of SSH protocol when closing channels.

louim:
I would love to get this merged, this fixed the error we had with Maverick_SSHD. The error looked like this:

```
E, [2025-05-14T21:28:39.720591 #12866] ERROR -- net.ssh.transport.session[20c0]: forcing connection closed
IOError: closed stream
from /Users/louismichel.couture/.local/share/mise/installs/ruby/3.4.3/lib/ruby/gems/3.4.0/gems/net-ssh-7.3.0/lib/net/ssh/transport/packet_stream.rb:74:in 'IO.select'
```

PR #679, opened 2019-06-19
Expose StrictHostKeyChecking from SSH config

### LINKED ISSUE #678: Honor StrictHostKeyChecking in SSH config
### Expected behavior

When connecting to a host that I don't have an host key signature for, I'd expect it to behave similar to SSH. Specifically, I've setup my `~/.ssh/config` with
```
Host *.mysite.example.com
  StrictHostKeyChecking no
  UserKnownHostsFile /dev/null
```
as these are ephemeral hosts that I may only connect to a few times. I can connect to them with `ssh host.mysite.example.com`. I would expect the same to work with Net::SSH without specific arguments.

I'd like the SSH connection to normally use fairly strict verification, unless SSH config is setup as above. Alternatively, I'd be interested in simple methods to set the `verify_host_key` option based on matching `StrictHostKeyChecking` in SSH config using `Net::SSH::Config.for(...)`.

### Actual behavior

Connection is rejected with
```
Host key verification failed for host.mysite.example.com: fingerprint SHA256:... is unknown for "host.mysite.example.com,10.0.0.2"
```

### System configuration

- net-ssh version 5.2.0
- Ruby version
ruby 2.5.3p105 (2018-10-18 revision 65156) [x86_64-darwin18]

### Example App

```ruby
gem 'net-ssh', '= 5.2.0'
require 'net/ssh'

@host = ENV['TEST_HOST']
@user = ENV['USER']
Net::SSH.start(@host, @user, check_host_ip: false, verify_host_key: :always) do |ssh|
end
```

Set 'TEST_HOST' to a target consistent with the SSH config above that you haven't added to your known_hosts file.

MikaelSmith:
Make this setting available so it can be used when deciding how to configure `verify_host_key` so that users of this library can enable behavior consistent with `ssh`.

Resolves #678.

### DIFF
--- a/lib/net/ssh/config.rb
+++ b/lib/net/ssh/config.rb
@@ -33,6 +33,7 @@ module Net
     # * ProxyJump => maps to the :proxy option
     # * PubKeyAuthentication => maps to the :auth_methods option
     # * RekeyLimit => :rekey_limit
+    # * StrictHostKeyChecking => :strict_host_key_checking
     # * User => :user
     # * UserKnownHostsFile => :user_known_hosts_file
     # * NumberOfPasswordPrompts => :number_of_password_prompts
@@ -202,6 +203,7 @@ module Net
             identityfile: :keys,
             fingerprinthash: :fingerprint_hash,
             port: :port,
+            stricthostkeychecking: :strict_host_key_checking,
             user: :user,
             userknownhostsfile: :user_known_hosts_file,
             checkhostip: :check_host_ip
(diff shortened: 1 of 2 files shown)


lucywyman:
@mfazekas Hello! This is a relatively small change - we know you're probably really busy, but let us know if there's anything we can do to help get it reviewed and merged. Anyone else we should ping? And IRC or slack channel to post to? Thanks!

MikaelSmith:
Thanks!

PR #681, opened 2019-07-05
Don't lookup host_keys in known_hosts if verify_host_key is set to :never

stillhart:
I don't think Net::SSH should lookup the host_key in the known_hosts files if verify_host_key is set to :never.
I had an issue when having a huge known_hosts file, that the parsing would take forever even if verify_host_key was set to :never. This is not what I expected. 

- Does openssh do the same?
- Are there any sideffects I am not considering?
- Test is missing


MikaelSmith:
It looks like some tests may have relied on that behavior and need to be updated (I assume that's why codecov dropped). Rubocop also prefers `&&` instead of `and`. Are those things you could update?

mfazekas:
The PR makes sense provided this is the openssh behavior as well

PR #682, opened 2019-07-11
Improve performance of known_hosts parsing

nicklewis:
This makes two changes to known_hosts parsing:
* Only go through the work of building a regex for pattern matching if
  the entry has pattern characters (* and ?). This was ~75% performance
  improvement in the case where no patterns are being used.
* Use simple string splitting instead of a string scanner when parsing a
  single line. This was ~20% performance improvement.

Together these changes entail ~80% reduction in parsing time with no
change in behavior. In a case with a large (20k entries) known_hosts
file, the time was reduced from 350ms to 65ms.

### DIFF
--- a/lib/net/ssh/known_hosts.rb
+++ b/lib/net/ssh/known_hosts.rb
@@ -130,27 +130,21 @@ module Net
         host_ip = entries[1]
 
         File.open(source) do |file|
-          scanner = StringScanner.new("")
           file.each_line do |line|
-            scanner.string = line
+            hosts, type, key_content = line.split(' ')
+            next unless hosts || hosts.start_with?('#')
 
-            scanner.skip(/\s*/)
-            next if scanner.match?(/$|#/)
+            hostlist = hosts.split(',')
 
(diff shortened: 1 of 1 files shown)


nicklewis:
This is of interest to us over in [Bolt land](https://github.com/puppetlabs/bolt), because our users run on dozens or hundreds of nodes at a time and the known_hosts file is parsed once for every host. That can add up tremendously.

There's potential for much more improvement here if it were changed to only parse the known_hosts file once and lookup keys as needed, but that's a larger structural change and would also be a behavior change (new keys added while running wouldn't be picked up). There's also some complexity around the ability to accept and add new keys to the known_hosts file.

A simpler step might be to provide a separate implementation of the KnownHosts class that has the desired behavior and just pass it in as `:known_hosts` in the options where needed. That way the slightly different behavior could be opt-in for scenarios where the caveats around updating it aren't a concern.

MikaelSmith:
Gotta love those code coverage changes. You removed some lines that were fully tested, so the percentage of tested lines dropped slightly.

MikaelSmith:
Anything else needed to merge this?

PR #686, opened 2019-07-31
Add cached KnownHosts implementation

nicklewis:
This adds a new class Net::SSH::KnownHosts::Cached which loads the
known_hosts files once, builds an index, and uses that when searching
for keys for a given host. An instance of this object can be passed
using the `known_hosts` option in order to avoid loading known_hosts
multiple times when making multiple connections to either the same host
or several hosts.

This behavior is implemented in a new class because it differs from the
existing implementation in the case where the content of the known_hosts
file changes outside the current process. With the standard
implementation, those changes will always be picked up the next time a
connection is made, whereas the Cached implementation won't
automatically incorporated changes made out-of-band.

### DIFF
--- a/lib/net/ssh/known_hosts/cached.rb
+++ b/lib/net/ssh/known_hosts/cached.rb
@@ -1,0 +1,106 @@
+module Net
+  module SSH
+    class KnownHosts
+      class Cached
+        def initialize(options)
+          @user_files = Array(options[:user_known_hosts_file] || %w[~/.ssh/known_hosts ~/.ssh/known_hosts2])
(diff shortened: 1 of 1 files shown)


nicklewis:
Opened this as a draft to get feedback on whether it's a good idea / desirable to have in net-ssh itself vs. implementing it in a separate library. If it's appropriate to incorporate to net-ssh, I'll get it finished up with docs and tests.

mfazekas:
@nicklewis thank for the PR, can't we check the file modification dates and reload cache if changed?

PR #688, opened 2019-08-12
Regenerate rubocop_todo.yml

fwininger:
Hello,

I would like to contribute some code to add `curve25519-sha256`, but before a submit this PR can you accept this one ? 

It's just a `rubocop --auto-gen-config` update.

Thanks


mfazekas:
@fwininger thanks is this needed because we don't have rubocop version fixed, and new rubocop has new checks?

fwininger:
I have use the rubocop version specify in the gemspec : 
```
spec.add_development_dependency "rubocop", "~> 0.54.0"
```

I can not update the rubocop version to the last, because the last version require ruby >= 2.3 and the Net::Ssh gem require ruby >= 2.2.6.

If you think that we can move forward, I will happy to do a second PR with the change of ruby requirement and the change the CI configuration to use the last ruby version.

fwininger:
@mfazekas please can you merge this PR ? I would like to use my summer time to do several enhancement to this gem. I will do another PR with the drop of ruby 2.2 and a update of developpement dependencies.

PR #685, opened 2019-07-31
Fix bundle 2.0 compatibility

fwininger:
This PR solve the bundle 2.0 compatibility issue : 

```
$ bundle install
Fetching gem metadata from https://rubygems.org/.............
Fetching gem metadata from https://rubygems.org/.
Resolving dependencies...
Bundler could not find compatible versions for gem "bundler":
  In Gemfile:
    bundler (~> 1.11)

  Current Bundler version:
    bundler (2.0.2)
This Gemfile requires a different version of Bundler.
Perhaps you need to update Bundler by running `gem install bundler`?

Could not find gem 'bundler (~> 1.11)' in any of the relevant sources:
  the local ruby installation
```

PR #677, opened 2019-06-06
Add support for all algorithms in Transport::Algorithms#host_key_format

brbrady:
Currently, only ssh-rsa is supported; others that also come in in the format `#{cipher_name}-cert-v0[01]@openssh.com` will break currently.

example error:
```
Net::SSH::Exception: host key algorithm mismatch for signature 'ecdsa-sha2-nistp256' != 'ecdsa-sha2-nistp256-cert-v01@openssh.com'
/var/lib/gems/2.3.0/gems/net-ssh-5.2.0/lib/net/ssh/transport/kex/ecdh_sha2_nistp256.rb:88:in `send_kexinit'
```

PR #515, opened 2017-04-11
Basic coding style with rubocop

fwininger:
This pull request fix the ruby coding for spaces.

PR #689, opened 2019-08-14
Remove trailing whitespaces

fwininger:
Fix every trailing whitespaces.


fwininger:
@mfazekas can you merge this trival PR ?

mfazekas:
@fwininger sorry i don't really like this change. I also don't like the whitespaces at the end of lines. 

But this change is a pain in the neck from git standpoint, it makes any merge before the change painfull, and it also makes a lot of noise in git history/annotate, i presonally don't think it's worth it.

fwininger:
I can only partially agree with you, I like having a clear git history but when the PR changes are mainly on empty lines, I think it's still to good move forward.

fwininger:
@mfazekas I plan to add modern algorithms to net/ssh : kex curve25519, ciphers aes-gcm, chacha20-poly1305. 
For curve25519 see #690, but new ciphers I need to refactor a lot of the source code, because AEAD encryptions don't need mac algorithms.
The current coding style is very crap and I think we really need a update to have a gem with better maintainability with new ruby versions.

Recently I have merged two PR :
- #685 : unlock the project to use with bundler 2.X. I did this because I can't install the gem in development mode.
- #688 : update rubocop_todo.yml to allow other contributer to don't commit ~500 lines of bullshit on every PR (see for example https://github.com/net-ssh/net-ssh/pull/677/files)

So I have two choices :
- I can fork the project and I do an new gem for the modern usage of ssh and allow other people to contribute easily
- I can merge progressively all my PR in net/ssh to have a better future maintainability .

Personally, I prefer the second option that follows my phiolosophy of the open source software, but if reject again my PR (see #515) I will choose the first option.

mfazekas:
@fwininger Sorry maybe i wasn't clear on reason of #515 rejection. It was rejected as net-ssh is a library that used by many popular apps, which means changes have to be reviewed so they doesn't contain any malware/lines with bad intentions.

Imagine a bad guy with bad intentions opening a PR with rubocop fixes and hiding 1 nasty lines along the trivial changes. See https://github.com/dominictarr/event-stream/issues/116 for example.

I know you don't have bad intentions but as a maintainer i still have to review changes coming from everyone. So you must understand i don't like PR-s with a lot of code changes, esp if they don't improve net-ssh functionality.

I'm very gratefull that you're working on net-ssh improvements. I'd like to ask your understanding and patience regarding huge number of code style changes.

As for this PR, regarding whitespaces as i see it:
+ it's an improvement as it removes unneccesary whitespaces
- it's a pain as it requires review in a lot of changes
- it's a pain as it creates conflits in unmerged commits, and complicates git blame/history (we don't have a huge number of commits, so that's also not a huge issue.)

But since it's important to you and you have non trivial PR #690, this is important for me as well. Please give me few days to try/review #690, and this PR then it can be merged, thans for your understanding.

fwininger:
Thanks for the answer, I understand your point with the need of review (as maintainer myself) and an the security concern (as Pentest myself).

Do you accept a PR with the requirement of ruby 2.3 in place of 2.2 ?

I can give you a tips to review this PR : 
```
git diff HEAD~1 --ignore-all-space --ignore-blank-lines
```

The output is null so I don't introduice malware code.

mfazekas:
Next version will be a major version so sure we can up required ruby version if needed. Some of our devops users, like to be able to support very old ruby versions but 2.3 sounds fine, it's already EOL.

Issue #476, opened 2016-12-29
net-ssh does not require ed25519 dependencies


mfazekas:
There are a number of (mostly natvie) gems required for ed25519 implementation to work.

- ```rbnacl-libsodium``` 
- ```rbnacl```
- ```bcrypt_pbkdf``` 

Currently it's up to the user to install them and net-ssh will complain, if those dependencies are not there/incorrect versions.

The following reasons led to this decision:

1. Net-ssh advertises itself as "Pure ruby implementation of (SSH"
2. bcrypt_pbkdf is not available on Java
3. rbnacl-libsodium doesn't have precompiled binaries for windows

mfazekas:
See capistrano/capistrano#1825

leehambley:
@mfazekas thanks for the numbered list, that clarifies the situation pretty well and confirms what I'd suspected after poking through the gemspec here and installation instructions for rbnacl.

I wonder if a solution here is to use a post-install message to clarify why this is the case, or whether there's a better solution?

mfazekas:
There is no current plan to solve this issue. The priority was getting 4.0 out of the door, and this compromise allows 4.0 to continue to run on wide range of platforms (eg Java and windows). https://github.com/cryptosphere/rbnacl-libsodium/issues/9 

There are the following options:

1. leave this as is and let those dependencies be someone else's issue. Improve documentation.
2. separate net-ssh to net-ssh-core and net-ssh where net-ssh-core is pure ruby, and net-ssh is the complete bundle with those dependencies marked as required. (or the other way around introduce a net-ssh-complete that has those dependencies)
3. add java support to bcrypt_pbkdf and add precompiled windows libs to rbnacl-libsodium and make net-ssh require them.
4. wait for new openssl to implement stuff needed for ed25519 support

leehambley:
> There is no current plan to solve this issue.

I'm not enthralled by any of the options set-out, but I'd imagine you aren't either. It's a real shame that we're no longer able to do SSH with pure Ruby. 

The "improve documentation" issue came up because we were deciding what to do with Capistrano, and SSHKit, and by then people are three layers removed from the source of the "problem".

Suggest the following: we treat this as the documentation issue that it is.

Quick question to that, is there a specific error raised when trying to connect to a server that expects `ed25519` keys, or is it a completely typical "Authentication Failure" ?

mattbrictson:
> Quick question to that, is there a specific error raised when trying to connect to a server that expects ed25519 keys

[...]

PR #690, opened 2019-08-14
Add Curve25519Sha256

fwininger:
Implementing the "curve25519-sha256@libssh.org" key-exchange algorithm. 

Defined in https://tools.ietf.org/html/draft-ietf-curdle-ssh-curves-06


mfazekas:
@fwininger thanks, looks great!!

Not really sure about the `x25519` dependency, with ed25516, we only use that if the user installed it, we might want to consider the same approach

See reasoning from ed25516 at:
https://github.com/net-ssh/net-ssh/issues/476#issue-197976226

And an integration test would be amazing!

fwininger:
I will do the correction quickly ! and after that I work on the implementation of new ciphers (aes-gcm and chacha20-poly1305).

Please can you merge my PR for coding style ? because I need to do some refactoring for the new ciphers.

fwininger:
Hello @mfazekas, I have rebase this PR and change the behaviour to load the code only when the gem x25516 is installed.

I have found no integration test example about kex.

To do advance test on my side I use docker with a debug compilation of openssh : https://github.com/fwininger/docker_ssh_debug

mfazekas:
Thanks! 🥇

PR #691, opened 2019-08-17
Update dependencies configuration

fwininger:
Follow up to #688, I update the dependencies configuration to use ruby 2.3.

Ruby 2.2 is no more supported and rubocop has also remove the support.


net-ssh.gemspec
```diff
@@ -33,10 +33,10 @@ Gem::Specification.new do |spec|
     spec.add_development_dependency("ed25519", "~> 1.2")
   end
 
-  spec.add_development_dependency "bundler", ">= 1.11"
+  spec.add_development_dependency "bundler", ">= 1.17"
 
   spec.add_development_dependency "minitest", "~> 5.10"
   spec.add_development_dependency "mocha", ">= 1.2.1"
   spec.add_development_dependency "rake", "~> 12.0"
-  spec.add_development_dependency "rubocop", "~> 0.54.0"
```
mfazekas: Note sure about freeeing rubucop. 

Doesn't it means that when a new rubocop is, the released CI can fail, with new rubocop checks.

fwininger: You have right ! I agree to lock the version to have no build error, the PR update 20 versions of rubocop, it good enough !

fwininger:
Can you merge this ? I have fix the rubocop version.

fwininger:
Thanks :)

PR #692, opened 2019-08-19
Consider ProxyCommand and ProxyJump together

zeos90:
ProxyCommand and ProxyJump override each other so they need to be
tracked together so the first one set takes precedence

### DIFF
--- a/lib/net/ssh/config.rb
+++ b/lib/net/ssh/config.rb
@@ -260,17 +260,8 @@ module Net
           when :preferredauthentications
             hash[:auth_methods] = value.split(/,/) # TODO we should place to preferred_auth_methods rather than auth_methods
           when :proxy
-            proxy_type, value = value
-
-            case proxy_type
-            when 'proxycommand'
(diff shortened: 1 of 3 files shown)


mfazekas (approved):

lib/net/ssh/config.rb
```diff
@@ -252,15 +259,20 @@ def translate_config_key(hash, key, value, settings)
...
+            proxy_type, value = value
+
+            case proxy_type
+            when 'proxycommand'
+              if value and value !~ /^none$/
+                require 'net/ssh/proxy/command'
+                hash[:proxy] = Net::SSH::Proxy::Command.new(value)
+              end
+            when 'proxyjump'
+              if value
```
mfazekas: I think this `if value` is unneccesary we're already checking if proxy_value is truthy when we add it.

zeos90: Good point, thanks. I'll remove them.

Issue #695, opened 2019-08-22
Modernize Openssl prerequirement


fwininger:
### Description

We have some piece of old code like this : 

```
if defined?(OpenSSL::Digest::SHA256)
  @digester = OpenSSL::Digest::SHA256
end
```

but other file use directly : 

```
def digester
  OpenSSL::Digest::SHA256
end
```

could it be possible to add a clear OpenSSL requirement to allow to clean code and make it more robust ?

I propose to add a prerequirement for OpenSSL 1.0.1 release in 2012.

### Expected behavior

- OpenSSL 1.0.0 [29 Mar 2010] introduce SHA2 (SHA256/SHA512)
- OpenSSL 1.0.1 [14 Mar 2012] introduce NIST-P224, NIST-P256, NIST-P521

### Actual behavior

In the readme we speak about : OpenSSL 0.9.8g 19 Oct 2007.

### Reference

https://www.openssl.org/news/changelog.html

fwininger:
@mfazekas what do you think about support only OpenSSL > 1.0.1 ?

Can I open a PR that enhance the code with this prerequirement ?

fwininger:
I have test with jruby 9.1 (ruby 2.3 compatible), OpenSSL::PKey::EC is defined. 

```sh
$ docker run -it jruby:9.1
irb(main):001:0> require 'openssl'
=> true
irb(main):002:0> OpenSSL::PKey::EC
=> OpenSSL::PKey::EC
```

https://github.com/jruby/jruby/wiki/Roadmap

mfazekas:
I'm fine with OpenSSL > 1.0.1

fwininger:
I have check aes-256-ctr is also introduice in OpenSSL 1.0.1.

https://security.stackexchange.com/questions/81243/openssl-with-aes-256-ctr-cipher/81308

I have try with Centos 6 / OpenSSL 1.0.1e 11 Feb 2013.

fwininger:
@mfazekas this issue can be close after #701 and #703.

PR #697, opened 2019-08-22
Remove unnecessary defined?(OpenSSL::PKey::EC) and defined?(OpenSSL::…

fwininger:
This PR change only test file and remove unnecessary defined?

Relative to #695


fwininger:
@mfazekas I have fix this PR, can you take a look ?

PR #699, opened 2019-08-23
Refactoring kex

fwininger:
- Move some code in the good place (Constant)
- Remove already include modules
- Define properly the get_p and get_g method to handle inheritance.


fwininger:
@mfazekas can you take a look ?

PS : I'm sorry but I still have a lot of PR in my backlog...

PR #701, opened 2019-08-23
Remove check because Openssl 1.0.1 introduice aes-ctr

fwininger:
relative with #695


fwininger:
@mfazekas can you merge that ?

PR #702, opened 2019-08-23
Refactoring kex

fwininger:
Enhance some code to clarify key exchange implemention.


fwininger:
I introduce a Astract class to kex to avoid DiffieHellmanGroup1SHA1 inhertance for EcdhSHA2NistP256 because the implementation of the KEX is not from the same RFC.

mfazekas:
Thanks looks much nicer!

Issue #705, opened 2019-08-24
Security issues


fwininger:
## Description

After a complet code audit, I found lot of critical security configurations.

I suggest to be OpenSSH 3.9/3.9p1 compatible, release on 2004-08-18, and remove all insecure and weak algorithme.

## Expected behavior

Be weak but not completly unsafe and backward compatible with OpenSSH 3.9/3.9p1.

Implement new algorithms.

## Actual behavior

Realy unsafe, allow communication without encryption !

### To remove

diffie-hellman-group1-sha1 : 
- removed (in server) since OpenSSH 6.7, unsafe algorithm
- disabled (in client) since OpenSSH 7.0, logjam attack
- using small 1024-bit modulus
- using weak hashing algorithm

diffie-hellman-group-exchange-sha1 :
- removed (in server) since OpenSSH 6.7, unsafe algorithm
- using weak hashing algorithm

### Should be remove

ecdh-sha2-nistp521 / ecdh-sha2-nistp384 / ecdh-sha2-nistp256 :
- using weak elliptic curves
- available since OpenSSH 5.7, Dropbear SSH 2013.62

NIST curves is unsecure see : https://safecurves.cr.yp.to/

Remove when curve25519-sha256@libssh is implement (see #690)

### To conserve

diffie-hellman-group14-sha1 :
- only for backward compatibility
- using weak hashing algorithm
- available since OpenSSH 3.9, Dropbear SSH 0.53

diffie-hellman-group-exchange-sha256 :
- only for backward compatibility
- using custom size modulus (possibly weak)
- available since OpenSSH 4.4

### To implement

curve25519-sha256@libssh.org :
- available since OpenSSH 6.5, Dropbear SSH 2013.62 (see #690)

### To remove

ssh-dss :
- removed (in server) and disabled (in client) since OpenSSH 7.0, weak algorithm
- using small 1024-bit modulus
- using weak random number generator could reveal the key
- available since OpenSSH 2.1.0, Dropbear SSH 0.28
                       
### Should be remove

ecdsa-sha2-nistp521 / ecdsa-sha2-nistp384 /ecdsa-sha2-nistp256
- using weak elliptic curves
- using weak random number generator could reveal the key

NIST curves is unsecure see : https://safecurves.cr.yp.to/

### To conserve

ssh-rsa :
- available since OpenSSH 2.5.0, Dropbear SSH 0.28

ssh-ed25519:
- available since OpenSSH 6.5

### To remove

[...]

PR #703, opened 2019-08-23
Modernize OpenSSL SHA2 implementation

fwininger:
relative #695

PR #706, opened 2019-08-24
Remove hmac-md5

fwininger:
Remove old hmac for security raisons.

See #705 

Merge after #703


fwininger:
Reject, see comment on #705

PR #707, opened 2019-08-27
Add changelog_uri to metadata to easily link from rubygems.org

nickhammond:
Reference: https://guides.rubygems.org/specification-reference/#metadata


mfazekas:
@nickhammond Thanks! Yep i guess the easiest is to remove spaces before `=`. The above line is not consitent either, alternative is to exclude this check for changelog. We

fwininger:
Yes, you can to `rubocop -a` to autocorrect this type of error.

Thanks for this PR.

PR #713, opened 2019-09-11
Integration test for Curve25519Sha256Keys

mfazekas:


mfazekas:
@fwininger FYI this is the integration test i for the kex:

It adds `KexAlgorithms curve25519-sha256` to sshd_config then if we can connect then we should be good.

fwininger:
Ok, understood for the example.

FYI : I will test the previous master branch (with only secure algs) on large scale (more 10000 than systems) next week. It's a little late to include curve25519, but I will test this on the 7th october. I'll tell you if we have anomalies.

fwininger:
hello @mfazekas, just a quick message to says that we have release the master branch of net::ssh on production with more than 10000 system without any error. :)

mfazekas:
> hello @mfazekas, just a quick message to says that we have release the master branch of net::ssh on production with more than 10000 system without any error. :)

@fwininger  thanks that sounds great! Many thanks for your contributions to the project.

I've just release net-ssh 6.0.0 beta1 to rubygems

Issue #441, opened 2016-10-20
Support more secure MAC algorithms


graaff:
Trying to secure our servers follow the feedback from https://github.com/arthepsy/ssh-audit I noticed that all current MAC algorithms in net-ssh are considered either insecure or suspect by ssh-audit. It would be great if additional MAC algorithms could be supported, specifically:

hmac-sha2-512-etm@openssh.com 
hmac-sha2-256-etm@openssh.com
umac-128-etm@openssh.com

All of these have been available in openssh since 6.2 so they should already be used widely.

fwininger:
#705

PR #714, opened 2019-09-15
Add sha2-{256,512}-etm@openssh.com MAC algorithms

graaff:
Implement the Encrypt-Then-Mac versions of the SHA2-256 and SHA2-512
MACs. These MACs are implemented by openssh and may be the only MACs
available on a hardened installation of openssh.

With EtM the MAC is calculated over the unencrypted packet length and
the encrypted payload (which includes padding length and padding).

The main benefit of EtM schemes is that it allows the encrypted
payload to be authenticated before it gets passed to the encryption
engine. This patch does not implement that mechanism, but this can be
added later to the poll_next_packet method. Note that all current MACs
already pass unauthenticated data to the encryption engine.

This implements part of #441


mfazekas:
@graaff thanks, looks great for the first glance, there are some rubocop issues, and we'll need some simple integration test too!

Great work! 👍

graaff:
I already had a look at the rubocop issues. I've fixed most of them and I guess I'll hide the others in the todo file? The CamelCase ones can't be fixed without breaking existing naming conventions, and I feel that things like the complexity and size counters are better dealt with in separate refactoring commits.

The integration tests seem to use tools that I have no experience with and that I don't have installed. I also don't see any integration tests for existing macs? Perhaps I am overlooking them. I've done manual integration tests against openssh 8.0.

Please let me know if you have other concerns or questions. I'll try to push another commit later this week.

fwininger:
Thanks!

related to #705

fwininger:
Can you update the Readme.md ?

lib/net/ssh/transport/packet_stream.rb
```diff
@@ -146,11 +146,32 @@ def enqueue_packet(payload)
...
-          mac = client.hmac.digest([client.sequence_number, unencrypted_data].pack("NA*"))
+          if client.hmac.etm
+            debug { "using encrypt-then-mac" }
 
-          encrypted_data = client.update_cipher(unencrypted_data) << client.final_cipher
-          message = encrypted_data + mac
+            # Encrypt padding_length, payload, and padding. Take MAC
+            # from the unencrypted packet_lenght and the encrypted
+            # data.
+            length_data = [packet_length].pack("N")
```
fwininger:
```suggestion
length_data = [packet_length].pack("N")
unencrypted_data = [padding_length, payload, padding].pack("CA*A*")
encrypted_data = client.update_cipher(unencrypted_data) << client.final_cipher

message = if client.hmac.etm
  # Encrypt padding_length, payload, and padding. Take MAC message = encrypted_data + mac
  # from the unencrypted packet_lenght and the encrypted data.
  mac_data = length_data + encrypted_data
  mac = client.hmac.digest([client.sequence_number, mac_data].pack("NA*"))
  mac_data + mac
else
  mac_data = length_data + unencrypted_data
  mac = client.hmac.digest([client.sequence_number, mac_data].pack("NA*"))
  encrypted_data + mac
end
```

test/transport/test_packet_stream.rb
```diff
@@ -978,6 +1098,9 @@ def test_enqueue_utf_8_packet_should_ensure_packet_length_is_in_bytes_and_multip
             srand(100)
             stream.client.set cipher: cipher, hmac: hmac, compression: compress
             stream.enqueue_packet(ssh_packet)
+            if PACKETS[cipher_name][hmac_name].nil? or PACKETS[cipher_name][hmac_name][compress].nil?
```
fwininger: I thinks you can remove this lines.

fwininger:
@graaff thanks you very much for this PR. I have do some work recently to implement the lastest security features in Net::SSH. 

After this PR, are you able to implement umac-128-etm@openssh.com in order to complet the HMAC section in #705 ?

Next step is to disable MAC with new encryption algs (aes-gcm and chacha20-poly1305)

test/transport/test_packet_stream.rb
graaff: Yes, that was debug code to help fill the packet test cases. Already gone in my local copy.

graaff:
I skipped umac-128-etm@openssh.com since umac is not part of openssl. A quick search doesn't seem to indicate a ruby implementation of that hash, so that would need to be included in the patch as well.

fwininger:
Ok perfect! We can skip umac-128-etm@openssh.com for now. The next step is to implement Authenticated encryption with associated data (AEAD) like aes-gcm or chacha20-poly1305.

mfazekas:
> The integration tests seem to use tools that I have no experience with and that I don't have installed. I also don't see any integration tests for existing macs? Perhaps I am overlooking them. I've done manual integration tests against openssh 8.0.

I'll add a simple one. Like adding `MACs hmac-sha2-512-etm@openssh.com` to `sshd_config` then see if we can connect.
Just like:  https://github.com/net-ssh/net-ssh/pull/713

lib/net/ssh/transport/packet_stream.rb
graaff: This change is not correct because it does not include the packet length in the encrypted data for the normal mac operations.  Fixing this would require bringing the update_cipher call into the if which is pretty much where my code was going, so I've kept my code.

graaff:
I think I applied all suggestions and remarks other than the refactoring suggestion and the integration tests. Rebased against current master and all tests are green.

mfazekas:
We're seeing some corrupted hmac detected in latest 6.0.0 - see #728. I'm going to release a 6.0.1 with etm opt-in, so this will be disabled by default - See #757

PR #720, opened 2019-10-24
Assume every string literal is frozen

mauriciopasquier:
Since ruby 2.3 we can use `RUBYOPT=--enable-frozen-string-literal` instead of magic comments (`# frozen_string_literal: true`) in every file, so it's not unusual to run the whole project with this setting enabled, for performance reasons. I think it's desirable to be frozen-friendly by default (also, frozen strings will be the default for ruby 3).

This commit dupes every string before its transformed. Maybe a naive approach, but it works.

I only had troubles in `test/transport/test_algorithms.rb`, I think because it was hitting exceptions, so I disabled the frozen strings explicitely there.


mfazekas:
> (also, frozen strings will be the default for ruby 3).

Not it won't be feault, see https://bugs.ruby-lang.org/issues/11473

kép

lib/net/ssh/buffer.rb
```diff
@@ -347,7 +347,7 @@ def write(*data)
       # Optimized version of write where the caller gives up ownership of string
       # to the method. This way we can mutate the string.
       def write_moved(string)
-        @content << string.force_encoding('BINARY')
+        @content << string.dup.force_encoding('BINARY')
```
mfazekas: see comment above the method, we don't want `dup` here for sure

mauriciopasquier: Duh, sorry. My bad

mauriciopasquier: I feel like it should fail a test then.

PR #722, opened 2019-10-29
Support :certkeys and CertificateFile configuration option

anderscarling:
Crude first implementation of supporting the CertificateFile ssh configuration option.

This is useful for using multiple key certificates with one identity file (e.g. when working with multiple organizations). Loading behaviour matches how OpenSSH ssh client handles the certificates, where an on-disk CertificateFile is able to match with a ssh-agent-based private key.

I've not had time to read up on how testing is organized yet, but will try to find some time to get some tests written soon. I would greatly appreciate comments on the implementation and if this is a feature you might want to merge (given appropriate test coverage etc).


anderscarling:
Tests are updated and written and should hopefully pass on CI now.

I ended up having to expand the testing in test_key_manager.rb a bit as the existing implementation stubbed out public certificates, which conflicted with the way certkeys where mapped to private certificates.

I think there is probably enough tests to prove that the implementation is working now, so I'll await feedback before I put more work into this PR.

anderscarling:
Is there anything I can do to ease the testing and (hopefully) inclusion of this?

It would be very valuable addition to net-ssh for me, and I would be willing to spend more time on a different or updated implementation if project maintainers see any problem with the current pull request.

mfazekas:
@anderscarling looks great to me, thanks for the PR! Sorry for the late response.

anderscarling:
No worries, just happy if we can get it merged. 

Please let me know if you think it needs any additional work!

mfazekas:
@anderscarling i've released net-ssh 6.0.0.beta2, thanks again the great PR!

anderscarling:
Awesome, thanks!

PR #727, opened 2019-11-17
Add ssh agent lock and unlock

Lapizistik:
Provide methods to use agent locking and unlocking with a password.

I am not sure how to test this properly.

### DIFF
--- a/lib/net/ssh/authentication/agent.rb
+++ b/lib/net/ssh/authentication/agent.rb
@@ -197,7 +197,7 @@ module Net
           raise AgentError, "could not lock agent" if type != SSH_AGENT_SUCCESS
         end
         
-        # unlock the ssh agend with password
+        # unlock the ssh agent with password
         def unlock(password)
           type, = send_and_wait(SSH2_AGENT_UNLOCK, :string, password)
           raise AgentError, "could not unlock agent" if type != SSH_AGENT_SUCCESS
--- a/lib/net/ssh/authentication/agent.rb
+++ b/lib/net/ssh/authentication/agent.rb
@@ -191,7 +191,7 @@ module Net
           raise AgentError, "could not remove all identity from agent" if type != SSH_AGENT_SUCCESS
         end
 
-        # lock the ssh agend with password
+        # lock the ssh agent with password
         def lock(password)
           type, = send_and_wait(SSH2_AGENT_LOCK, :string, password)
           raise AgentError, "could not lock agent" if type != SSH_AGENT_SUCCESS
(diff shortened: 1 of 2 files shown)


lib/net/ssh/authentication/agent.rb
```diff
@@ -189,6 +191,18 @@ def remove_all_identities
           raise AgentError, "could not remove all identity from agent" if type != SSH_AGENT_SUCCESS
         end
 
+        # lock the ssh agend with password
```
mfazekas, line 194:
```suggestion
        # lock the ssh agent with password
```

lib/net/ssh/authentication/agent.rb
```diff
@@ -189,6 +191,18 @@ def remove_all_identities
...
+        def lock(password)
+          type, = send_and_wait(SSH2_AGENT_LOCK, :string, password)
+          raise AgentError, "could not lock agent" if type != SSH_AGENT_SUCCESS
+        end
+        
+        # unlock the ssh agend with password
```
mfazekas, line 200:
```suggestion
        # unlock the ssh agent with password
```

mfazekas:
@Lapizistik thanks for the PR! Can you document the usecase, and a sample session?!


https://tools.ietf.org/html/draft-miller-ssh-agent-02#section-4.6

Lapizistik:
> thanks for the PR! Can you document the usecase, and a sample session?!

I am not sure how to proceed from here. Can you give me some pointers where the documentation should go (and how use case descriptions/samples of other ssh agent sessions are described)?

My actual usecase is a rather complicated setup and will not work well as an example use case. An example use case could be to be able to lock your agent (i.e. your key to the git server) before you invite someone for pair programming with byobu on your development VM.

The use case I am working on (and for which I did this patch) involves a fileserver and two backup servers, where the internal backup-server wakes up at night, pulls its backup from the fileserver, then tells the fileserver to start the external backup and finally goes back to sleep while the fileserver opens the ssh connection to the external server and uses restic to do the backup.
And it simply does not feel good to have the ssh-agent around with the open credentials all the time, able to modify/delete my external backup. So the server is locked and the password for unlocking is provided from the internal backup server.
I totally understand that this does not help against an experienced attacker who has gained access to my fileserver: they can simply wait for the moment the ssh-server gets unlocked (or intercept the password from teh backup server). 
But many attacks I currently see are automated or semi-automated (e.g. cryptoviral ransomware) or done by not-so-much-experienced hackers, and here locking the ssh-agent helps.

mfazekas:
@Lapizistik thanks for the explanation, sure the use cases you documented makes sense.

There is a rubocop error to be fixed:
```
$ bundle _1.17_ exec rubocop

Inspecting 171 files

...C.......................................................................................................................................................................

Offenses:

lib/net/ssh/authentication/agent.rb:283:1: C: Layout/TrailingBlankLines: 1 trailing blank lines detected.

171 files inspected, 1 offense detected

The command "bundle _1.17_ exec rubocop" exited with 1.
```

PR #730, opened 2019-11-20
Fix unexpected IOError

christianrolle:
Connectiong with a remote Windows system ends up in unexpected
IOErrors, which let the connection fail totally.
The dedicated location, where the IOError happens couldn't be
detected exactly. Though it should be rescued.


mfazekas:
@christianrolle thanks for the PR!
To tell the truth i'm not fan of silencing errors without understanding what's the root cause. Is this something you can reliably reproduce?!

christianrolle:
@mfazekas The remote system is the [PSFTPd Secure FTP Server](https://www.psftp.de/ftp-server/) by Microsoft. We can not reliably reproduce the IO exceptions. But they happen from time to time.

PR #744, opened 2020-03-01
Add cached KnownHosts implementation - continued...

donoghuc:
This PR continues the work that @nicklewis did in https://github.com/net-ssh/net-ssh/pull/686 based on the suggestion from @mfazekas to rebuild cache when file changes occur out of band. There are two approaches for this in the PR. First use file mtime (originally suggested by @mfazekas) and the second to use an md5 hash to determine if a known_hosts file has changed out of band. While writing tests for the original mtime implementation I came across a case where a file write happened so close to the time the file was originally read that comparing mtimes did not detect a change.

### DIFF
--- a/lib/net/ssh/known_hosts/cached.rb
+++ b/lib/net/ssh/known_hosts/cached.rb
@@ -1,3 +1,5 @@
+require 'digest/sha2'
+
 module Net
   module SSH
     class KnownHosts
@@ -6,11 +8,13 @@ module Net
       # to determine what key a user prefers to use for a given host. This is different
       # From the KnownHosts class in that it builds a cache of parsed known hosts in order
       # to avoid re-parsing known hosts earch time it is needed.
+      #
+      # An instance of this class can optionaly be used by setting the :known_hosts option.
       class Cached
         def initialize(options)
           @user_files = Array(options[:user_known_hosts_file] || %w[~/.ssh/known_hosts ~/.ssh/known_hosts2])
           @global_files = Array(options[:global_known_hosts_file] || %w[/etc/ssh/ssh_known_hosts /etc/ssh/ssh_known_hosts2])
-          @mtimes = {}
+          @sha_sums = {}
         end
 
         def search_for(host, options={})
@@ -45,7 +49,7 @@ module Net
           return unless File.readable?(source)
 
           File.open(source) do |file|
-            @mtimes[source] = File.mtime(source)
+            @sha_sums[source] = Digest::SHA256.hexdigest(File.read(file))
             file.each_line do |line|
               hosts, type, key_content = line.split(' ')
               # Skip empty line or one that is commented
@@ -86,10 +90,10 @@ module Net
 
           @user_files.each do |file|
             begin
-              # If this is the only modification since last read, update @mtimes to preserve cache
-              preserve_cache = File.mtime(file) == @mtimes[file]
+              # If this is the only modification since last read, update @sha_sums to preserve cache
+              preserve_cache = File.mtime(file) == @sha_sums[file]
               KnownHosts.new(file).add(host, key)
-              @mtimes[file] = File.mtime(file) if preserve_cache
+              @sha_sums[file] = Digest::SHA256.hexdigest(File.read(file)) if preserve_cache
               break
             rescue SystemCallError
             end
@@ -109,7 +113,13 @@ module Net
 
         def cache_invalid?
           (@user_files + @global_files).each do |file|
-            return true if File.readable?(file) && @mtimes[file] != File.mtime(file)
+            # Checksum is different, rebuild
+            if File.readable?(file) && @sha_sums[file] != Digest::SHA256.hexdigest(File.read(file))
+              return true
+            # File has been deleted, rebuild
+            elsif @sha_sums[file] && !File.readable?(file)
+              return true
+            end
           end
           false
         end
--- a/lib/net/ssh/known_hosts/cached.rb
+++ b/lib/net/ssh/known_hosts/cached.rb
@@ -13,8 +13,13 @@ module Net
       class Cached
         def initialize(options)
           @user_files = Array(options[:user_known_hosts_file] || %w[~/.ssh/known_hosts ~/.ssh/known_hosts2])
+          @user_files.map! { |file| File.expand_path(file) }
           @global_files = Array(options[:global_known_hosts_file] || %w[/etc/ssh/ssh_known_hosts /etc/ssh/ssh_known_hosts2])
+          @global_files.map! { |file| File.expand_path(file) }
           @sha_sums = {}
+          @host_lookups = {}
+          @hmac_lookups = {}
+          @pattern_lookups = {}
         end
 
         def search_for(host, options={})
--- a/lib/net/ssh/known_hosts/cached.rb
+++ b/lib/net/ssh/known_hosts/cached.rb
@@ -1,0 +1,106 @@
+module Net
+  module SSH
+    class KnownHosts
+      class Cached
+        def initialize(options)
+          @user_files = Array(options[:user_known_hosts_file] || %w[~/.ssh/known_hosts ~/.ssh/known_hosts2])
+          @global_files = Array(options[:global_known_hosts_file] || %w[/etc/ssh/ssh_known_hosts /etc/ssh/ssh_known_hosts2])
+          build_cache
+        end
+
+        def search_for(host, options={})
+          hostname, host_ip = host.split(',')
+
+          entries = []
+          entries.concat @host_lookups.fetch(hostname, [])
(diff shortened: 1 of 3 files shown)


lib/net/ssh/known_hosts/cached.rb
```diff
@@ -17,12 +24,10 @@ def search_for(host, options={})
           @hmac_lookups.each do |(hmac, salt), entry|
             digest = OpenSSL::Digest.new('sha1')
             host_hmac = OpenSSL::HMAC.digest(digest, salt, hostname)
-            entries << entry if Base64.encode64(host_hmac.chomp) == hmac
+            entries.concat(entry) if Base64.encode64(host_hmac).chomp == hmac
```
donoghuc: This was a bug whereby using << would insert an array rather than merging.

lib/net/ssh/known_hosts/cached.rb
```diff
@@ -40,9 +45,11 @@ def parse_known_hosts(source)
           return unless File.readable?(source)
 
           File.open(source) do |file|
+            @mtimes[source] = File.mtime(source)
             file.each_line do |line|
               hosts, type, key_content = line.split(' ')
-              next unless hosts || hosts.start_with?('#')
```
donoghuc: Pulls in work from https://github.com/net-ssh/net-ssh/pull/742

lib/net/ssh/known_hosts/cached.rb
```diff
@@ -54,7 +61,7 @@ def parse_known_hosts(source)
               hostlist.each do |host|
                 entry = Entry.new(hostlist, key)
                 if host.include?('*') || host.include?('?')
-                  regex = regexify_pattern(host)
```
donoghuc: I think this was just a typo.

lib/net/ssh/known_hosts/cached.rb
```diff
@@ -73,14 +80,17 @@ def parse_known_hosts(source)
           end
         end
 
-        def add(host, key)
+        def add(host, key, _options = nil)
```
donoghuc: I needed to keep support for a third positional parameter for `Hostskeys#add_host_key` https://github.com/net-ssh/net-ssh/blob/54476fa1e88a046014fa986792dbf2c65c5f17db/lib/net/ssh/known_hosts.rb#L24

test/test_known_hosts_cached.rb.rb
```diff
@@ -0,0 +1,208 @@
...
+
+  def test_host_entry_added_out_of_process
+    Tempfile.open('github') do |f|
+      options = { user_known_hosts_file: f.path }
+      kh = Net::SSH::KnownHosts::Cached.new(build_opts(f.path))
+      keys = kh.search_for('github.com', options)
+      assert_equal(0, keys.count)
+      f.write(File.read(path("known_hosts/github")))
+      f.close
+      # Ensure mtime is different
```
donoghuc: This is where I found issues with mtimes being so close that the file would not be re-parsed.

lib/net/ssh/known_hosts/cached.rb
```diff
@@ -0,0 +1,127 @@
...
+          (@user_files + @global_files).each do |file|
+            parse_known_hosts(File.expand_path(file))
+          end
+        end
+
+        def parse_known_hosts(source)
+          return unless File.readable?(source)
+
+          File.open(source) do |file|
+            @md5_sums[source] = Digest::MD5.hexdigest(File.read(file))
```
fwininger: Please use SHA256 instead of MD5.

lib/net/ssh/known_hosts/cached.rb
```diff
@@ -0,0 +1,127 @@
...
+          false
+        end
+      end
+
+      Entry = Struct.new(:hosts, :key)
+    end
+  end
+end
+
+
```
fwininger: remove one line

test/test_known_hosts_cached.rb.rb
```diff
@@ -0,0 +1,202 @@
...
+      assert_equal(0, keys.count)
+      assert_equal([], keys.to_a)
+
+      kh.add('github2.com',rsa_key)
+      keys2 = kh.search_for("github2.com")
+      assert_equal([rsa_key.to_blob], keys2.to_a.map(&:to_blob))
+    end
+  end
+
+  def
```
fwininger: remove this

donoghuc: Wow, yeah that was an interesting mistake. That was effectively causing that test to be skipped. I'm surprised that did not cause any warnings or errors. :man_shrugging: fixed.

lib/net/ssh/known_hosts/cached.rb
donoghuc: This seems like an easy change to make. For now i'd like to see what folks think about using a hash instead of mtime. If we want to move forward with hash I think using sha256 should be just fine.

lib/net/ssh/known_hosts/cached.rb
```diff
@@ -0,0 +1,125 @@
...
+              Regexp.escape(y)
+            end.join('.')
+          end.join('[^.]*')
+
+          Regexp.new("\\A#{pattern_regexp}\\z")
+        end
+
+        def cache_invalid?
+          (@user_files + @global_files).each do |file|
+            return true if File.readable?(file) && @md5_sums[file] != Digest::MD5.hexdigest(File.read(file))
```
mfazekas: I think it also needs to check if the file is removed, that should invalidate cache as well

mfazekas:
>  and the second to use an md5 hash to determine if a known_hosts file has changed out of band.

Yes i think using some hash is a great idea, much better than `mtime`!

lib/net/ssh/known_hosts/cached.rb
donoghuc: Updated to used sha256

lib/net/ssh/known_hosts/cached.rb
donoghuc: Updated to invalidate on delete.

donoghuc:
@nicklewis Did you have any scripts to test performance with this change?

donoghuc:
Looked a bit at what gains there would be from using cached known_hosts. I built up a very large known hosts file with the following strategy: Take an existing known_hosts entry and save everything but the host name. Build up 100K host entries with a `www.foo_[1..100k],1.2.3.4 ssh-ed25519 ....` then in the last line add the actual "good" known_hosts entry. 

## Cached
Test script for using cached:
```ruby
#!/usr/bin/env ruby

require 'net/ssh'
require 'net/ssh/known_hosts/cached'
options = {
  verify_host_key: Net::SSH::Verifiers::Always.new,
  keys: '/home/cas/.ssh/id_rsa-acceptance',
  known_hosts: Net::SSH::KnownHosts::Cached.new({})
}

(0..100).each do 
  Net::SSH.start('tasty-calibre.delivery.puppetlabs.net', 'root', options)
end
```
Example time:
```
real	0m42.614s
user	0m8.878s
sys	0m1.049s
```
Profile: https://gist.github.com/donoghuc/0e13a59b544935f1b87316cbbc2b4f03#file-cached_hosts-svg

## Not cached
Script for non-cached:
```ruby
#!/usr/bin/env ruby

require 'net/ssh'
require 'net/ssh/known_hosts/cached'
options = {
  verify_host_key: Net::SSH::Verifiers::Always.new,
  keys: '/home/cas/.ssh/id_rsa-acceptance'
}

(0..100).each do 
  Net::SSH.start('tasty-calibre.delivery.puppetlabs.net', 'root', options)
end
```
Time example:
```
real	0m55.168s
user	0m21.868s
sys	0m0.446s
```
Profile: https://gist.github.com/donoghuc/0e13a59b544935f1b87316cbbc2b4f03#file-non_cached_hosts-svg

I also noticed that sha256 is a bit slower than md5: Given collisions seem like they would be extremely rare could we just use md5 (or some even faster algorithm) as there are not really any security concerns, just detecting file changes?
## sha256
```ruby
#!/usr/bin/env ruby
require 'digest/sha2'
path = File.expand_path('~/.ssh/known_hosts')
(0..1000).each do 
  Digest::SHA256.hexdigest(File.read(path))
end
```
```
real	0m33.535s
user	0m29.791s
sys	0m3.723s
```
## md5
```ruby
#!/usr/bin/env ruby
require 'digest/md5'
path = File.expand_path('~/.ssh/known_hosts')
(0..1000).each do 
  Digest::MD5.hexdigest(File.read(path))
end
```
```
real	0m22.207s
user	0m18.901s
sys	0m3.307s
```

logicminds:
How would this work with centralized based knownhosts database implementations?

ie. https://gitlab.com/nwops/net-ssh-exthosts/-/tree/master

PR #745, opened 2020-03-09
Add set_env option

### LINKED ISSUE #725: Feature request: add a new ssh option `set_env`, in addition to `send_env`
### Expected behavior

Currently, this gem only supports forwarding environment variables in ENV class. In a process lifecycle, when we want to start multiple ssh sessions concurrently, each with different value for a same environment variable name, `send_env` will not help.

I'm using Capistrano to deploy rails application to servers that's behind a bastion server, and I need to pass server name and other information as ssh environment variable to bastion server to select which server. For `config/deploy/production.rb`,

```
ENV['AssetName'] = 'server1'
ENV['SystemUserName'] = 'jenkins'
server '192.168.1.2', user: 'jenkins', ssh_options: { send_env: ['AssetName', 'SystemUserName'], keys: File.expand_path('~/.ssh/jenkins.pem') }

# Here I cannot deploy to more than one server concurrently, because of `send_env` limitation.
# ENV['AssetName'] = 'server2'
# ENV['SystemUserName'] = 'jenkins'
# server '192.168.1.2', user: 'jenkins', ssh_options: { send_env: ['AssetName', 'SystemUserName'], keys: File.expand_path('~/.ssh/jenkins.pem') }
```

Capistrano depends on sshkit which depends on this gem. Of course I can make suggestion to sshkit team, but I think sshkit just passes ssh_options to `Net::SSH.start`. So maybe it make more senses to suggest such change in this gem.

I suggest adding a new option `set_env`, the usage is like this:

```
Net::SSH.start('host', 'user', set_env: { AssetName: 'server1', SystemUserName: 'jenkins' })
Net::SSH.start('host', 'user', set_env: { AssetName: 'server2', SystemUserName: 'jenkins' })
```

It makes sense that `send_env` and `set_env` can both exist. `send_env` specifies which names from ENV you want to expose, and `set_env` specifies which pairs of environment name and value you want to expose. And for a same name, `set_env` will win against `send_env`.

I know that we can use `Net::SSH::Connection::Channel#env` to add environment variables, but it will be more straightforward to pass a hash of environment variables as an option of ssh. It will also match more with what linux command ssh supports: `ssh -o "SetEnv=AssetName=server1 SystemUserName=jenkins"`

What do you think of this? If you think this is a good idea, I can make a PR to add such feature.

uzxmx:
Closes #725.

This PR adds a new SSH option `set_env` that is supported in [OpenSSH 7.8](https://www.openssh.com/txt/release-7.8). For more info, please see #725.

### DIFF
--- a/lib/net/ssh/connection/channel.rb
+++ b/lib/net/ssh/connection/channel.rb
@@ -2,8 +2,8 @@ require 'net/ssh/loggable'
 require 'net/ssh/connection/constants'
 require 'net/ssh/connection/term'
 
-module Net 
-  module SSH 
+module Net
+  module SSH
     module Connection
 
       # The channel abstraction. Multiple "channels" can be multiplexed onto a
@@ -530,6 +530,7 @@ module Net
           @remote_maximum_packet_size = max_packet
           connection.forward.agent(self) if connection.options[:forward_agent] && type == "session"
           forward_local_env(connection.options[:send_env]) if connection.options[:send_env]
+          set_remote_env(connection.options[:set_env]) if connection.options[:set_env]
           @on_confirm_open.call(self) if @on_confirm_open
         end
     
@@ -676,6 +677,13 @@ module Net
               self.env(env_name, env_value)
             end
           end
+        end
+
+        # Set a +Hash+ of environment variables in the remote process' environment.
+        #
+        #   channel.set_remote_env foo: 'bar', baz: 'whale'
+        def set_remote_env(env)
+          env.each { |key, value| self.env(key, value) }
         end
       end
 
--- a/lib/net/ssh/connection/channel.rb
+++ b/lib/net/ssh/connection/channel.rb
@@ -682,9 +682,12 @@ module Net
         # Set a +Hash+ of environment variables in the remote process' environment.
         #
         #   channel.set_remote_env foo: 'bar', baz: 'whale'
+        #
+        # rubocop:disable Naming/AccessorMethodName
         def set_remote_env(env)
           env.each { |key, value| self.env(key, value) }
         end
+        # rubocop:enable Naming/AccessorMethodName
       end
 
     end
--- a/lib/net/ssh/connection/channel.rb
+++ b/lib/net/ssh/connection/channel.rb
@@ -682,12 +682,9 @@ module Net
         # Set a +Hash+ of environment variables in the remote process' environment.
         #
         #   channel.set_remote_env foo: 'bar', baz: 'whale'
-        #
-        # rubocop:disable Naming/AccessorMethodName
         def set_remote_env(env)
           env.each { |key, value| self.env(key, value) }
         end
-        # rubocop:enable Naming/AccessorMethodName
       end
 
     end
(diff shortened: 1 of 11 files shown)


lib/net/ssh/connection/channel.rb
```diff
@@ -677,6 +678,16 @@ def forward_local_env(env_variable_patterns)
             end
           end
         end
+
+        # Set a +Hash+ of environment variables in the remote process' environment.
+        #
+        #   channel.set_remote_env foo: 'bar', baz: 'whale'
+        #
+        # rubocop:disable Naming/AccessorMethodName
```
mfazekas: pls just disable Naming/AccessorMethodName from config

uzxmx: Got it. Meanwhile, I'm trying to add some integration test for what I've done, and upgrade OpenSSH used in integration test from 7.4p1 to 7.9p1. Because I saw the codecov/patch check was not passed, seemed the test coverage was not enough. Will let you have a look after I finish.

mfazekas: It's complining that `set_remote_env` in channel.rb is not tested you might get away without integation test but testing that line, but sure an integration would be great!

test/integration/common.rb
```diff
@@ -94,6 +94,7 @@ def with_lines_as_tempfile(lines = [], &block)
 
   # @yield [pid, port]
   def start_sshd_7_or_later(port = '2200', config: nil)
+    pid = nil
```
uzxmx: Without pre-declaring `pid`, when `config` exists, the `pid` within `with_lines_as_tempfile` block will be out of scope, which results that statements in `ensure` part doesn't execute as expected (in other words, sshd process won't be killed).

test/integration/test_channel.rb
```diff
@@ -51,11 +51,11 @@ def test_transport_close_before_channel_close_should_raise
       proxy = Net::SSH::Proxy::Command.new("/bin/nc localhost 22")
       res = nil
       Net::SSH.start(*ssh_start_params(proxy: proxy)) do |ssh|
-        chanell_success_handler = lambda do
+        channel_success_handler = lambda do
```
uzxmx: Fix typo

test/integration/test_forward.rb
```diff
@@ -24,18 +24,6 @@
 class ForwardTestBase < NetSSHTest
   include IntegrationTestHelpers
 
-  # @yield [pid, port]
-  def start_sshd_7_or_later(port = '2200')
```
uzxmx: Remove this method because it's already been implemented in `IntegrationTestHelpers` that is mixed in.

uzxmx:
@mfazekas I've disabled rubocop offenses from config. Also I've added an integration test case for new `set_env` option. When I implemented that, I found a trivial bug and some typos, also removed a duplicated method definition. Detailed information is added as inline comment.

test/integration/common.rb
mfazekas: good catch thanks!

PR #746, opened 2020-03-11
Limit cert.valid_before to max supported by JRuby

sskousen:
The OpenSSH certificate format interpets a `valid_before` time of `2^64-1` (max for unsigned long) to mean "forever": https://github.com/openssh/openssh-portable/blob/master/sshkey.c#L3159

Golang also interprets it this way:
https://github.com/golang/crypto/blob/master/ssh/certs.go#L47

I'm running into a `RangeError` issue with net-ssh on JRuby for certificates that have been issued to be valid for 'forever', as JRuby uses a signed long, and also multiplies the passed value by 1000 (as the internal storage uses milliseconds). (https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/RubyTime.java#L1280)

This caps the `cert.valid_before` at `0x20c49ba5e353f7` which is the largest value that can be handled by JRuby and which translates to `292278994-08-17 01:12:55 -0600`, which I believe is close enough to 'forever' to not have any real effect here.

### DIFF
--- a/lib/net/ssh/authentication/certificate.rb
+++ b/lib/net/ssh/authentication/certificate.rb
@@ -34,11 +34,11 @@ module Net
           # 0x20c49ba5e353f7 = 0x7fffffffffffffff/1000, the largest value possible for JRuby
           # JRuby Time.at multiplies the arg by 1000, and then stores it in a signed long.
           # 0x20c49ba5e353f7 = 292278994-08-17 01:12:55 -0600
-          if RUBY_PLATFORM == "java"
-            cert.valid_before = Time.at([0x20c49ba5e353f7, buffer.read_int64].min)
-          else
-            cert.valid_before = Time.at(buffer.read_int64)
-          end
+          cert.valid_before = if RUBY_PLATFORM == "java"
+                                Time.at([0x20c49ba5e353f7, buffer.read_int64].min)
+                              else
+                                Time.at(buffer.read_int64)
+                              end
 
           cert.critical_options = read_options(buffer)
           cert.extensions = read_options(buffer)
--- a/lib/net/ssh/authentication/certificate.rb
+++ b/lib/net/ssh/authentication/certificate.rb
@@ -34,7 +34,12 @@ module Net
           # 0x20c49ba5e353f7 = 0x7fffffffffffffff/1000, the largest value possible for JRuby
           # JRuby Time.at multiplies the arg by 1000, and then stores it in a signed long.
           # 0x20c49ba5e353f7 = 292278994-08-17 01:12:55 -0600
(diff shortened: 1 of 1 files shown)


lib/net/ssh/authentication/certificate.rb
```diff
@@ -31,7 +31,10 @@ def self.read_certblob(buffer, type)
           cert.key_id = buffer.read_string
           cert.valid_principals = buffer.read_buffer.read_all(&:read_string)
           cert.valid_after = Time.at(buffer.read_int64)
-          cert.valid_before = Time.at(buffer.read_int64)
+          # 0x20c49ba5e353f7 = 0x7fffffffffffffff/1000, the largest value possible for JRuby
+          # JRuby Time.at multiplies the arg by 1000, and then stores it in a signed long.
+          # 0x20c49ba5e353f7 = 292278994-08-17 01:12:55 -0600
+          cert.valid_before = Time.at([0x20c49ba5e353f7, buffer.read_int64].min)
```
mfazekas: Please put it to an if statement so only executed on jruby

lib/net/ssh/authentication/certificate.rb
```diff
@@ -34,7 +34,12 @@ def self.read_certblob(buffer, type)
           # 0x20c49ba5e353f7 = 0x7fffffffffffffff/1000, the largest value possible for JRuby
           # JRuby Time.at multiplies the arg by 1000, and then stores it in a signed long.
           # 0x20c49ba5e353f7 = 292278994-08-17 01:12:55 -0600
-          cert.valid_before = Time.at([0x20c49ba5e353f7, buffer.read_int64].min)
+          if RUBY_PLATFORM == "java"
```
mfazekas: sorry to be picky, but can you please put comment also inside if. 
I generally don't like comments as they often live a separate life from the code, keeping them closer together might help.

https://thecodinglove.com/code-comments

lib/net/ssh/authentication/certificate.rb
sskousen: I debated if it was better to have consistent behavior on all platforms, or limit to just JRuby. Fixed.

lib/net/ssh/authentication/certificate.rb
sskousen: Great idea. Fixed. I also don't like comments like this, but decided that it was needed to explain the magic number of `0x20c49ba5e353f7`.

sskousen:
@mfazekas Thank you for merging this request. Will I need to wait for net-ssh 6.0 for this to be released, or do you plan on backporting it to 5.2?

PR #751, opened 2020-04-08
Support algorithm subtraction syntax from ssh_config

maxfierke:
As per `man ssh_config` for KexAlgorithms, MACs, and similar options:

```
Multiple algorithms must be comma-separated. Alternately if the specified value begins with a '+' character, then the specified methods will be appended to the default set instead of replacing them.  If the specified value begins with a '-' character, then the specified methods (including wildcards) will be removed from the default set instead of replacing them.
```

Without this, having these subtraction options in your SSH config will replace the default algorithms and cause Net::SSH to raise with `Net::SSH::Exception` and "could not settle on host_key algorithm". i.e. If you set `HostKeyAlgorithms -ssh-rsa`, it uses `-ssh-rsa` as the algorithm, rather than removing `ssh-rsa` from the algorithm preferences.

### DIFF
--- a/lib/net/ssh/transport/algorithms.rb
+++ b/lib/net/ssh/transport/algorithms.rb
@@ -290,10 +290,24 @@ module Net
           list = []
           option = Array(option).compact.uniq
 
-          if option.first && option.first.start_with?('+')
+          if option.first && option.first.start_with?('+', '-')
             list = supported.dup
-            list << option.first[1..-1]
-            list.concat(option[1..-1])
+
+            appends = option.select { |opt| opt.start_with?('+') }.map { |opt| opt[1..-1] }
+            deletions = option.select { |opt| opt.start_with?('-') }.map { |opt| opt[1..-1] }
+
+            list.concat(appends)
+
+            deletions.each do |opt|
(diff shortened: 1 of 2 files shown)


maxfierke:
AppVeyor CI is failing on a 404 for the jRuby version configured. Unrelated to PR.

lib/net/ssh/transport/algorithms.rb
```diff
@@ -290,10 +290,26 @@ def compose_algorithm_list(supported, option, append_all_supported_algorithms =
...
             list = supported.dup
-            list << option.first[1..-1]
-            list.concat(option[1..-1])
+
+            appends = option.select { |opt| opt.start_with?('+') }.map { |opt| opt[1..-1] }
+            deletions = option.select { |opt| opt.start_with?('-') }.map { |opt| opt[1..-1] }
+
+            list.concat(appends) if appends.any?
+
+            if deletions.any?
```
fwininger, line 301: remove any? you can iterate on a empty array

maxfierke: ah yeah, good call. done in c457c06

Issue #728, opened 2019-11-18
Exception : corrupted hmac detected


fwininger:
### Actual behavior

In some case, I have this error :
- corrupted hmac detected

### System configuration

- net-ssh master
- Ruby 2.6.1

### related issue 

Other people seems to have also this kind of problem :
https://github.com/ytti/oxidized/issues/1952

fwininger:
Did you have any idea to solve this kind of issue ?

mfazekas:
@fwininger first step would be some kind of somewhat reliable way to reproduce. 
Then I'd try to add some logging to figure out what can be the issue.

But it doesn't sound trivial at all.

stillhart:
I experienced this issue too using net-ssh and I can reproduce it on our equipment. In my case I am using https://github.com/swisscom/net-ssh-cli. It only happens when I run certain commands on my machines. @fwininger What devices are you connecting to using oxidized?

I will try to reproduce this using only net-ssh.

fwininger:
I have no idea to reproduce this error systematically, but I use directly Net::SSH over 10000 VMs.

stillhart:
I could reproduce the error using plain net-ssh on our equipment. It happens after a random amount of tries but it does happen after a while.
https://gist.github.com/stillhart/70785e3e1ee8dc3f43e0200c0e22f5de

The error happens here when receiving a few hundret our thousand lines of text:
```ruby
channel.send_data "display event history all list\n"
until output[/#\z/] do
   ssh.process(0.000001) 
end
```

Does anybody have an idea how to debug this further? How can I find out if the device is the issue or net-ssh?

mfazekas:
@stillhart one thing i'd try is logging a lot of stuff at the place we throw the exception:

https://github.com/net-ssh/net-ssh/blob/5c31c9dc5f65b151a9195d1d7c307e2f6d155da1/lib/net/ssh/transport/packet_stream.rb#L251-L266

So i'd do something like this at line l263:
```ruby
if real_hmac != my_computed_hmac 
  puts "Real hmac: #{real_hmac.inspect}"
  puts "My hmac: #{my_computed_hmac}"
  puts "packet_length: #{@packet_length} padding_length: #{padding_length}"
  puts "server.hmac.mac_length: #{server.hmac.mac_length}"
  puts "server.hmac.etm :#{server.hmac.etm}"
  raise Net::SSH::Exception, "corrupted hmac detected"
end
```

then that might give us more information on what's going on.

[...]

PR #759, opened 2020-04-25
Fix + test for corrupted hmac detected issue

mfazekas:
Fixes: #728


fwininger:
I have a test environnement with 200 vm that I can use next week to validate that there a no more issue.

mfazekas:
> I have a test environnement with 200 vm that I can use next week to validate that there a no more issue.

Thanks  much, i'll release 6.0.2 with this bug etm fixed but still opt-in. And will release 6.1.0-rc1 with etm default.

PR #570, opened 2018-02-05
Use OpenSSL for aes*-ctr

mfazekas:
Closes #542

PR #767, opened 2020-05-27
Re-add fix for JRuby

sskousen:
This re-adds the changes that were introduced in https://github.com/net-ssh/net-ssh/pull/612 and removed in https://github.com/net-ssh/net-ssh/pull/701.

While OpenSSL since 2012 has support for `aes-ctr`, the latest JRuby (9.2.11.1) still does not support them, which means that net-ssh 6 is broken for anyone using JRuby and has `aes-ctr` enabled on the server:

```
$ rbenv shell jruby-9.2.11.1
$ irb
irb(main):001:0> require 'openssl'
=> true
irb(main):002:0> ::OpenSSL::Cipher.ciphers.include?("aes-256-ctr")
=> false
irb(main):003:0> ::OpenSSL::Cipher.ciphers.include?("aes-192-ctr")
=> false
irb(main):004:0> ::OpenSSL::Cipher.ciphers.include?("aes-128-ctr")
```

### DIFF
--- a/lib/net/ssh/transport/cipher_factory.rb
+++ b/lib/net/ssh/transport/cipher_factory.rb
@@ -22,9 +22,9 @@ module Net
           "3des-ctr"                    => "des-ede3",
           "blowfish-ctr"                => "bf-ecb",
 
-          "aes256-ctr"                  => ::OpenSSL::Cipher.ciphers.include?("aes-256-ctr") ? "aes-256-ctr" : "aes-256-cbc",
-          "aes192-ctr"                  => ::OpenSSL::Cipher.ciphers.include?("aes-192-ctr") ? "aes-192-ctr" : "aes-192-cbc",
-          "aes128-ctr"                  => ::OpenSSL::Cipher.ciphers.include?("aes-128-ctr") ? "aes-128-ctr" : "aes-128-cbc",
+          "aes256-ctr"                  => ::OpenSSL::Cipher.ciphers.include?("aes-256-ctr") ? "aes-256-ctr" : "aes-256-ecb",
+          "aes192-ctr"                  => ::OpenSSL::Cipher.ciphers.include?("aes-192-ctr") ? "aes-192-ctr" : "aes-192-ecb",
+          "aes128-ctr"                  => ::OpenSSL::Cipher.ciphers.include?("aes-128-ctr") ? "aes-128-ctr" : "aes-128-ecb",
           'cast128-ctr'                 => 'cast5-ecb',
 
           'none'                        => 'none'
--- a/lib/net/ssh/transport/cipher_factory.rb
+++ b/lib/net/ssh/transport/cipher_factory.rb
@@ -22,9 +22,9 @@ module Net
           "3des-ctr"                    => "des-ede3",
           "blowfish-ctr"                => "bf-ecb",
 
-          'aes256-ctr'                  => 'aes-256-ctr',
-          'aes192-ctr'                  => 'aes-192-ctr',
-          'aes128-ctr'                  => 'aes-128-ctr',
+          "aes256-ctr"                  => ::OpenSSL::Cipher.ciphers.include?("aes-256-ctr") ? "aes-256-ctr" : "aes-256-cbc",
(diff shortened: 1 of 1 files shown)


fwininger:
I'm not sure if it's the best idea, AES-CTR is the only secure cipher in 2020 we have.

Temporalily, I think you should specify the cipher list you want as Net::SSH params.

mfazekas:
@fwininger I think this still means `AES-CTR` just we'll not use native openssl primitive with ctr, but we extend them to support ctr.

This used to be the deafault impl, and in net-ssh 5.0 it was a perf opt to use native `aes-*-ctr` instead of `aes-*-cbc` + `Net::SSH::Transport::CTR` #570 


https://github.com/net-ssh/net-ssh/blob/74919ef69344164bd55e2dfc7ee2a360b6c718c6/lib/net/ssh/transport/cipher_factory.rb#L55-L61

sskousen:
@fwininger I agree, which is why I have AES-CTR enabled on all my boxes (and first cipher on the list). I believe this is a reasonable middle ground, as CTR is still preferred if available, but net-ssh will fall back to CBC if CTR is not enabled. 

Without something like this, a stock net-ssh install on JRuby connecting to any stock Ubuntu server (and I would assume any other modern SSH server) will fail with a cryptic 'padding' error (that took me 2 days to figure out what it was causing it). That seems like an extremely bad user experience.

Ideally, I think net-ssh should query the underlying openssl library for supported modes and ciphers, disable insecure ones, and then use those in the initial SSH negotiation. Then instead of a padding error, there would be an error about negotiation failing because they couldn't find a common cipher, which is a lot closer to the root problem. I don't know enough ruby, the internals of net-ssh, or the ssh protocol to be able to build that, though I'd be willing it give it a try if pointed in the right direction.

I can work on explicitly setting the ciphers, but we're using an internal tool built on top of capistrano 2, which is built on top of net-ssh, so there's several layers that I'll have to dig through to try and get that setup correctly. Any user of JRuby will also have to do something similar.

mfazekas:
>  I agree, which is why I have AES-CTR enabled on all my boxes (and first cipher on the list). I believe this is a reasonable middle ground, as CTR is still preferred if available, but net-ssh will fall back to CBC if CTR is not enabled.

My understading is that this is not failing back to CBC, but uses ruby code to add CTR instead of doing it all in openssl. See https://github.com/net-ssh/net-ssh/pull/767#issuecomment-635279279

sskousen:
@mfazekas If I'm reading #570 correctly, it appears that we're implementing CTR in ruby on top of OpenSSL's ECB mode, is that correct (and very awesome). I thought this change was simply switching which AES model would be used and so changed it to CBC, since ECB is horribly insecure. I will need to change it back to ECB to have CTR working again.

sskousen:
👍 I was just a bit to slow in commenting. That makes sense, and I didn't know it was possible to implement CTR on top of ECB. I'll update the MR back to ECB.

mfazekas:
> If I'm reading #570 correctly, it appears that we're implementing CTR in ruby on top of OpenSSL's ECB mode, is that correct (and very awesome).

Yep, that is where we started from pre 5.0. Then with 5.0 we optimized to use CTR from OpenSSL as it was much faster than CRuby but that was broken on OpenSSL. But that is not working with JRuby, so we're just using the Ruby CTR again on jruby.

sskousen:
@mfazekas This PR is currently configured to use CBC instead of ECB (if CTR isn't available). Do I need to switch that back to ECB?

mfazekas:
> @mfazekas This PR is currently configured to use CBC instead of ECB (if CTR isn't available). Do I need to switch that back to ECB?

`¯\_(ツ)_/¯`, if you don't have strong opinion then i'd suggest ECB as that was working fine in `5.0.1+`. In any case you should be able to test by connecting ot a server where this is the only option.

sskousen:
Switched back to ECB.

thbar:
Thank you everyone who worked on this - I had issues with SFTP & SSH connections from JRuby when upgrading `net/ssh` from `5.2.0` to `6.1.0`,  and the latest `6.2.0.rc1` fixes this apparently. Thanks!

PR #781, opened 2020-08-06
Fix missing OpenSSL::Pkey::EC::Point#ssh_signature_type

haccht:
`OpenSSL::Pkey::EC::Point#ssh_signature_type` is missing.  
It should be an alias for the method `OpenSSL::Pkey::EC::Point#ssh_type`.

### DIFF
--- a/lib/net/ssh/transport/openssl.rb
+++ b/lib/net/ssh/transport/openssl.rb
@@ -53,9 +53,7 @@ module OpenSSL
         "ssh-rsa"
       end
 
-      def ssh_signature_type
-        ssh_type
-      end
+      alias ssh_signature_type ssh_type
 
       # Converts the key to a blob, according to the SSH2 protocol.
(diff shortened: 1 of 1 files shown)


fwininger:
👍

lib/net/ssh/transport/openssl.rb
```diff
@@ -245,6 +245,10 @@ def ssh_type
           "ecdsa-sha2-#{CurveNameAliasInv[group.curve_name]}"
         end
 
+        def ssh_signature_type
```
donoghuc, line 248:
```suggestion
alias ssh_signature_type ssh_type
````
It might be more clear to just use ruby's `alias` keyword for this.

haccht: Changed the code according to your suggestion. Also changed other `ssh_signature_type` methods as well.

PR #784, opened 2020-09-10
Fix uncommon blocking condition in version negotiation

shamblesides:
We're seeing issues where the version negotiation occasionally blocks indefinitely when communicating with an unstable SFTP server, even when timeout is specified. This attempts to fix the problem.

### DIFF
--- a/lib/net/ssh/transport/server_version.rb
+++ b/lib/net/ssh/transport/server_version.rb
@@ -46,15 +46,25 @@ module Net
           socket.write "#{PROTO_VERSION}\r\n"
           socket.flush
     
-          raise Net::SSH::ConnectionTimeout, "timeout during server version negotiating" if timeout && !IO.select([socket], nil, nil, timeout)
           loop do
             @version = ""
             loop do
               begin
-                b = socket.readpartial(1)
+                if timeout
+                  b = socket.read_nonblock(1)
+                else
+                  b = socket.readpartial(1)
+                end
                 raise Net::SSH::Disconnect, "connection closed by remote host" if b.nil?
               rescue EOFError
                 raise Net::SSH::Disconnect, "connection closed by remote host"
+              rescue IO::WaitReadable => e
+                timed_out = IO.select([socket], nil, nil, timeout).nil?
+                if timed_out
+                  raise Net::SSH::ConnectionTimeout, "timeout during server version negotiating"
+                else
+                  retry
+                end
               end
               @version << b
               break if b == "\n"
--- a/lib/net/ssh/transport/server_version.rb
+++ b/lib/net/ssh/transport/server_version.rb
@@ -50,15 +50,11 @@ module Net
             @version = ""
             loop do
               begin
-                if timeout
-                  b = socket.read_nonblock(1)
-                else
-                  b = socket.readpartial(1)
-                end
+                b = socket.read_nonblock(1)
                 raise Net::SSH::Disconnect, "connection closed by remote host" if b.nil?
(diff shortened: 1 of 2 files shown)


mfazekas:
@shamblesides Thanks much for the PR!

Can you tell us more about what's the issue is exactly and how this is working around the issue?! Also is this something we can reproduce, what is the SFTP server in question?!

shamblesides:
Hello!

We're uploading a lot of files to a lot of different third-party SFTP clients simultaneously. One of our clients' server seems to, very rarely (< 1/100 uploads) stop. We've turned on verbose debugging and the logs stop around here:

```
D, [2020-09-04T18:26:25.484125 #6599] DEBUG -- net.ssh.transport.session[2aede]: establishing connection to <<<address omitted>>>
D, [2020-09-04T18:26:25.505272 #6599] DEBUG -- net.ssh.transport.session[2aede]: connection established
I, [2020-09-04T18:26:25.505466 #6599]  INFO -- net.ssh.transport.server_version[2aee8]: negotiating protocol version
D, [2020-09-04T18:26:25.505566 #6599] DEBUG -- net.ssh.transport.server_version[2aee8]: local is `SSH-2.0-Ruby/Net::SSH_6.1.0 java'
```

and then it stops, and blocks forever. Preferably, since I specified a timeout argument, I would like the job to eventually fail so it will automatically retry.

I don't know the details about what SFTP server they are using. I was able to make the code fail in the same way by taking a Node SFTP server and editing this line of this dependency, so it doesn't send over the version correctly. https://github.com/mscdex/ssh2-streams/blob/master/lib/ssh.js#L241

However it might be possible that it's breaking on the socket.write call above where I made my changes. I'm not exactly sure. I just would like to avoid any potential untimed blocking in this code.

bschmeck:
I've run into this "indefinite hang" issue when (mistakenly) connecting to an FTP server, instead of an SFTP server.  In those cases, the remote is never going to send a version string (or any data) and we wind up blocking forever.  I've been running this branch in production for a few months now without any trouble.  Is there anything I can help with to get this merged into master?

shamblesides:
I’m no longer working at the company that was using this, paging @ETetzlaff @film42  ?

film42:
We would still like to see this merged.

steerlink:
Linked Issue: https://github.com/net-ssh/net-ssh/issues/788

I'm taking this over from @shamblesides since my team still has access to the SFTP endpoints, environment and source code.

After extensive investigation, the team was unable to reproduce an indefinite block during version exchange. However, we did find a blocked call to a forked process which was waiting for stdin.write. Discovered Net::SSH::Disconnect was getting raised and rescued in a path that left the forked process waiting for .write or .close.

The 1/1000 metric was likely caused by an anti-flood hammer running on the SFTP endpoint coupled with the size of the transfer volume per connection. It was a rare thing to connect often enough to trigger a Denial of Service.

Real issue has been resolved internally.

Recommend closing.

bschmeck:
I can still recreate an indefinite hang by connecting to a server that doesn't send a version string but also doesn't close the connection.  In practice, this is generally due to connecting to an FTP server accidentally, but it can be recreated by just running netcat `nc -l -p 4444` and then trying to connect via `Net::SSH.start("localhost", "user", password: "password", port: 4444)`

steerlink:
Gotcha. Since this PR has conflicts, would you be willing to open a new PR @bschmeck?

bschmeck:
No problem.  The new PR is #900

PR #796, opened 2021-01-22
Fix crash when an empty payload is received

alexander-lazarov:
Have fixed a crash whenever a packet with an empty payload is received:

```
unexpected response  (#<Net::SSH::Packet:0x00007f9d9fd83940 @named_elements={}, @content="", @position=0, @type=nil>)
```

The crash occurs when trying to transfer data via SFTP, the remote server is`SSH-2.0-CerberusFTPServer_8.0`. Unfortunately it is a private server and I cannot provide access for debugging. I'm happy to help with debugging if further investigation is needed.

### DIFF
--- a/lib/net/ssh/transport/packet_stream.rb
+++ b/lib/net/ssh/transport/packet_stream.rb
@@ -271,7 +271,7 @@ module Net
           server.increment(@packet_length)
           @packet = nil
 
-          return Packet.new(payload)
+          return payload.empty? ? nil : Packet.new(payload)
         end
       end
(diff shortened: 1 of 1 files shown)


mfazekas:
@alexander-lazarov thank for the PR, any reason it's addressed here and not on `net-sftp` side?!
I'm not 100% that empty packets can be just ignored for all net-ssh use cases.

alexander-lazarov:
Sorry, completely missed the reply here! I no longer have access to the SFTP server that was causing the problem, so I'll close my PR.

PR #799, opened 2021-02-16
Update agent.rb

Fguhhla:
Add excetion handle to read_paket because of the following error while using net-ssh 6.1 with test-kitchen 

from /opt/chef-workstation/embedded/lib/ruby/gems/2.7.0/gems/net-ssh-6.1.0/lib/net/ssh/authentication/agent.rb:211:in read_packet' /opt/chef-workstation/embedded/lib/ruby/gems/2.7.0/gems/net-ssh-6.1.0/lib/net/ssh/buffer.rb:147:in append': can't modify frozen String: "" (FrozenError)

### DIFF
--- a/lib/net/ssh/authentication/agent.rb
+++ b/lib/net/ssh/authentication/agent.rb
@@ -222,8 +222,13 @@ module Net
         # is returned as a Net::SSH::Buffer).
         def read_packet
           buffer = Net::SSH::Buffer.new(@socket.read(4))
-          buffer.append(@socket.read(buffer.read_long))
-          type = buffer.read_byte
+          begin
+            buffer.append(@socket.read(buffer.read_long))
+          rescue StandardError => e  
+            puts e.message  
+            puts e.backtrace.inspect  
+          end  
+            type = buffer.read_byte
           debug { "received agent packet #{type} len #{buffer.length - 4}" }
           return type, buffer
         end
--- a/lib/net/ssh/authentication/agent.rb
+++ b/lib/net/ssh/authentication/agent.rb
@@ -228,7 +228,7 @@ module Net
             puts e.message  
             puts e.backtrace.inspect  
           end  
-            type = buffer.read_byte
+          type = buffer.read_byte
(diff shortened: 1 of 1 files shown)


mfazekas:
@Fguhhla thanks much for the PR, I'd be more interested in the reason for the crash vs such a workaround, first it silences errors, second it prints to stdout which might not be appropriate for all apps.

mfazekas:
Isn't it possible that the code was ran In all strings frozen mode?! 
RUBYOPT="--enable-frozen-string-literal"

lib/net/ssh/authentication/agent.rb
```diff
@@ -222,7 +222,12 @@ def send_packet(type, *args)
         # is returned as a Net::SSH::Buffer).
         def read_packet
           buffer = Net::SSH::Buffer.new(@socket.read(4))
-          buffer.append(@socket.read(buffer.read_long))
+          begin
+            buffer.append(@socket.read(buffer.read_long))
+          rescue StandardError => e  
+            puts e.message
```
fwininger, line 228: Ignore StandardError here is a very bad choice.

fwininger (changes requested):

tas50:
@mfazekas We freeze strings throughout Test Kitchen

mfazekas:
@tas50 can you elaborate?! Do you use `enable-frozen-string-literal` globally ?!  `net-ssh` doesn't support this at the moment. [Buffer](https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/buffer.rb) for sure is not frozen literal compatible.

tas50:
You know it looks like we're not actually doing it in 100% of the files like I thought. We do set the frozen string literal magic comment in a large number of the Test Kitchen files though since it has a nice memory impact and at one point the world was prepping for that as the default in Ruby 3.

mfazekas:
@Fguhhla thanks much for the PR. We'd need to understand why this error happens, just silencing it is not solution. 

I'm closing this PR, let us know if you have more background info on this.

PR #801, opened 2021-03-11
Moved password =nil to inside password loop and update version

deligatedgeek:


deligatedgeek:
My git-fu caused me to create this PR prematurely, I need to do some tidy up and parameter passing.

deligatedgeek:
Need to handle banner prompts during device authentication in Oxidized, that led me to modifying net-ssh.
This does work, I just need to provide options to pass parameters for

1. Whether to automatically agree with any non password banners
2. A user defined regex for password prompt (if this isnt already being passed)
3. Potentially a regex for banner prompt

PR #802, opened 2021-03-12
Handle banner prompt within authentication step using keyboard-interactive

deligatedgeek:
We use Oxidized and have an issue due by a customer login banner requirement that prompts for agreement before the password prompt. Net-ssh doesn't currently implement this.

The banner prompt is relatively standard now and uses ssh message types 60 and 61 under keyboard-interactive. I have amended the keyboard-interactive authentication code to check the prompt, and if it is not asking for the password, give a 'yes' response or respond with the correct password.


mfazekas:
See https://github.com/net-ssh/net-ssh/issues/804#issuecomment-798836390

Issue #712, opened 2019-09-11
Add support for rsa-sha2-256, rsa-sha2-512


zzambers:
Please add support for rsa-sha2-256, rsa-sha2-512 algorithms for rsa keys (required for ssh server on RHEL-8 in FIPS mode). It should be possible use existing rsa-sha keys ( openssh client can do that ), and this RFC also says so: [1].

Background:
We had problem with Vagrant (which uses net-ssh library) not being able to connect to RHEL-8 VM running in FIPS mode. Log of OpenSSH server showed following lines:
...
Sep 11 13:00:47 localhost.localdomain sshd[2125]: Using arbitrary primes is not allowed in FIPS mode. Falling back to known groups.
Sep 11 13:00:48 localhost.localdomain sshd[2125]: userauth_pubkey: key type ssh-rsa not in PubkeyAcceptedKeyTypes [preauth]
Sep 11 13:00:48 localhost.localdomain sshd[2125]: Connection closed by authenticating user tester 192.168.121.1 port 55486 [preauth]
...

There was no such problem when connecting using openssh client. ( Net-ssh worked only with password auth. ) After some digging I found that net-ssh seems to only support rsa-sha, while openssh server on FIPS enabled RHEL-8 system requires rsa-sha2-256 or rsa-sha2-512 for rsa keys.

Here is how sshd command looks like as started by systemd on FIPS enabled RHEL-8 machine:

[...]

PR #811, opened 2021-05-06
Added support for RSA client authentication with SHA-2

zzambers:
Hi,

This PR adds support for client authentication using RSA keys together with SHA-2 [1] (rsa-sha2-256 / rsa-sha2-512). I have made issue for this some time ago [2]. Issue was closed with this PR [3], but client authentication was not added.

Issue used to affect only RHEL-8 FIPS mode (and was workaroundable), but now it became more serious as it also affects Fedora 34. RSA keys for client authentication do not work on f34 even with DEFAULT crypto policy (using latest net-ssh library).

I decided to take a look into code and try to fix this myself. With this change net-ssh tries to use rsa-sha2-256 in case ssh-rsa method fails using RSA key. This change fixes problem for me.

[1] https://tools.ietf.org/html/rfc8332
[2] https://github.com/net-ssh/net-ssh/issues/712
[3] https://github.com/net-ssh/net-ssh/pull/771


lib/net/ssh/authentication/methods/publickey.rb
```diff
@@ -89,6 +91,22 @@ def authenticate_with(identity, next_service, username)
...
           end
+
+          def authenticate_with(identity, next_service, username)
+            alg = identity.ssh_type
+            salg = nil
+            if authenticate_with_2(identity, next_service, username, alg, salg)
+              return true
+            end
+            if !rsa_sha2_auth_disable && alg == "ssh-rsa"
+              # if ssh-rsa fails, retry with sha2
```
Jakuje, line 102: The RSA with SHA2 should be used by default, not only if the SHA1 fails. This is just a very small subset of RFC 8332. Hardcoding the SHA256 regardless what is negotiated can also cause interoperability issues. 

https://datatracker.ietf.org/doc/html/rfc8332

lib/net/ssh/authentication/methods/publickey.rb
```diff
@@ -27,41 +27,43 @@ def authenticate(next_service, username, password=nil)
...
           end
 
           # Attempts to perform public-key authentication for the given
           # username, with the given identity (public key). Returns +true+ if
           # successful, or +false+ otherwise.
-          def authenticate_with(identity, next_service, username)
+          def authenticate_with_2(identity, next_service, username, alg, salg)
             debug { "trying publickey (#{identity.fingerprint})" }
-            send_request(identity, username, next_service)
+            send_request(identity, username, next_service, alg)
```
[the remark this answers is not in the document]
zzambers: This should be correct as it is, because value of ssh_type (not ssh_signature_type) is passed here, which preserves existing behaviour. See more here:
https://github.com/net-ssh/net-ssh/pull/811#discussion_r629628505

lib/net/ssh/authentication/key_manager.rb
```diff
@@ -171,8 +171,14 @@ def sign(identity, data)
           end
 
           if info[:key]
-            return Net::SSH::Buffer.from(:string, identity.ssh_signature_type,
-              :mstring, info[:key].ssh_do_sign(data.to_s)).to_s
+            if salg.nil?
+                signed = info[:key].ssh_do_sign(data.to_s);
```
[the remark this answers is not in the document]
zzambers: True, I should probably use better name.

lib/net/ssh/authentication/methods/publickey.rb
zzambers: I think certificates should be just ignored currently (SHA2 is not tried for them) as check for RSA is done by comparing ssh_type to "ssh-rsa" and this should fail for certs:
https://github.com/net-ssh/net-ssh/blob/c6a21e5f0a9f0bc92e9c4eb0e17af40d175a97a5/lib/net/ssh/authentication/certificate.rb#L53

zzambers: Do I understand it correctly that userauth request message is the same, just certificate is used instead of key?

Jakuje: In that case, using SHA256 is probably fine. I probably forgot about that part of RFC :)

The certificate authentication is using the same messages with different key types. So right, for now, the will be ignored, but you might also add something like this to support them, I think:

```
            if !rsa_sha2_auth_disable && alg == "ssh-rsa-cert-v01@openssh.com"
              # if ssh-rsa certificate fails, retry with sha2
              alg = salg = "rsa-sha2-256-cert-v01@openssh.com"
              return authenticate_with_2(identity, next_service, username,
                alg, salg)
            end
```

lib/net/ssh/authentication/methods/abstract.rb
```diff
@@ -21,12 +21,16 @@ class Abstract
           # this.
           attr_reader :key_manager
 
+          # Disable rsa SHA-2 auth, usefull for tests
+          attr_reader :rsa_sha2_auth_disable
```
fwininger, line 25: I'm not agree to had this, if it's only for units test. The best way is to disable in the test suite on server side. By the ways, you need to add more test on the PR.

berlin4apk:
Is something missing / unsolved that blocks the merge of this PR? @fwininger

zzambers:
I have done these improvements:
- Code style fixes (so robocop is not complaining)
- RSA algorithms to try can now be specified, also rsa-sha2-* are now tried first
- Support for  rsa-sha2-* certificates has been added
- Support for  rsa-sha2-* signatures, using ssh agent has been added

zzambers:
I performed manual testing with RSA keys/certs connecting to ssh server from: 
- RHEL-6
- RHEL-7
- RHEL-8+fips
- Fedora 34

In those configurations:
- private key using file
- private key using agent
- certificate using file
- certificate using agent

All worked

zzambers:
I also tried to use DSA keys to connect to RHEL-6 an RHEL-7 ssh server, just to make sure non-RSA keys work. Also worked.

zzambers:
Note:
Option pubkey_algorithms is currently internal (and only used by tests) and currently only affects RSA algorithms. However in the future it can be used/extended to implement list of enabled pubkey algorithms, which is not yet implemented. I have seen something like that already exists for e.g host key algorithms [1] (even reading of HostKeyAlgorithms ssh_config file option seems supported [2]), but there is currently no support for list of enabled pubkey algorithms (and PubkeyAcceptedAlgorithms option [3]). However, support for this, I think, is outside of scope of this PR. 

[1] https://github.com/net-ssh/net-ssh/blob/a0d62413d04abc9f3fbc32b55dee51c973ff693b/lib/net/ssh/transport/algorithms.rb#L27
[2] https://github.com/net-ssh/net-ssh/blob/a0d62413d04abc9f3fbc32b55dee51c973ff693b/lib/net/ssh/config.rb#L22
[3] https://man7.org/linux/man-pages/man5/ssh_config.5.html

zzambers:
I have just added some tests and improved naming of some variables and functions a bit.

I think I have now addressed all issues I could.

GabrielNagy:
We started to encounter this issue as well with RedHat 8 FIPS with Puppet, confirmed that this PR fixes the issue. Thanks @zzambers!

@fwininger, @mfazekas, can we help with anything to get this PR merged and released?

mfazekas:
@mfazekas if you can try this branch and confirm that it works fine for you that's a big help

ciprianbadescu:
@mfazekas, was the above message intended for you?

GabrielNagy:
@mfazekas I have tried this branch with RedHat 8 FIPS and confirm it works.

zzambers:
I am closing this PR, since rebased version of it got merged, See: https://github.com/net-ssh/net-ssh/pull/838
Thanks

PR #817, opened 2021-06-10
Fix unit tests

fwininger:


mfazekas (commented):
Thanks! I think we need to move away from travis to GitHub actions ASAP. Travis is no longer automatically free for open source project you have to ask for credits which they gave me after a few months. But if we've used the credits we've have to ask again.
Unfortunately I've had some issue running integrations tests on GitHub actions.

Issue #613, opened 2018-06-21
Support cert based host auth


habnabit:
### Expected behavior

When connecting to a host, if the host publishes only an `ssh-rsa` host key and `~/.ssh/known_hosts` contains an `ecdsa-sha2-nistp256` fingerprint for the host, OpenSSH will ignore the known host key and verify the `ssh-rsa` host key. In this case, despite the `ssh-rsa` key not being known, it is signed by a known host and considered trusted. I expected net-ssh to behave the same.

### Actual behavior

net-ssh does not trust the host, because it found _some_ known key, even if it's not of a key type the host is still publishing: https://github.com/net-ssh/net-ssh/blob/29a8e97ae8d2eaf674a8694b775dc119985fa5e9/lib/net/ssh/verifiers/always.rb#L17-L35

```
/Users/habnabit/.rvm/gems/ruby-2.3.1/gems/net-ssh-5.0.2/lib/net/ssh/verifiers/always.rb:46:in `process_cache_miss': fingerprint SHA256:[redacted] does not match for "[redacted]" (Net::SSH::HostKeyMismatch)
	/Users/habnabit/.rvm/gems/ruby-2.3.1/gems/net-ssh-5.0.2/lib/net/ssh/verifiers/always.rb:32:in `verify'
	/Users/habnabit/.rvm/gems/ruby-2.3.1/gems/net-ssh-5.0.2/lib/net/ssh/verifiers/accept_new.rb:18:in `verify'
	/Users/habnabit/.rvm/gems/ruby-2.3.1/gems/net-ssh-5.0.2/lib/net/ssh/verifiers/accept_new_or_local_tunnel.rb:17:in `verify'
	/Users/habnabit/.rvm/gems/ruby-2.3.1/gems/net-ssh-5.0.2/lib/net/ssh/transport/kex/diffie_hellman_group1_sha1.rb:188:in `verify_server_key'
	/Users/habnabit/.rvm/gems/ruby-2.3.1/gems/net-ssh-5.0.2/lib/net/ssh/transport/kex/diffie_hellman_group1_sha1.rb:72:in `exchange_keys'
	/Users/habnabit/.rvm/gems/ruby-2.3.1/gems/net-ssh-5.0.2/lib/net/ssh/transport/algorithms.rb:389:in `exchange_keys'
[...]

PR #833, opened 2021-08-05
Cert based host auth

mfazekas:
Fixes: #613 

Implements `@cert-authority` marker in KnownHosts file.
1. If we see such marker, matching the connected host, we allow prioritize all cert-algs when negotiating with server.
https://github.com/net-ssh/net-ssh/blob/51dc78feac5dc9df615ca0ef82e707951887411c/lib/net/ssh/known_hosts.rb#L41-L47
2. If server sends a certificate , we verify that it's signed correctly and the signature matches what's in known hosts
https://github.com/net-ssh/net-ssh/blob/51dc78feac5dc9df615ca0ef82e707951887411c/lib/net/ssh/known_hosts.rb#L54-L60

Also changes `*` in `known_host` like in openssh this now includes dots and also a single `*` can be used to match any hosts.

TODO: We don't support `@revoke` yet, also if nonstandard port like `2213` used, the we only match`[hostname]:2213` against patterns in `known_hosts`. Unlike regular opens which matches portless name if host key's is new. 
https://github.com/openssh/openssh-portable/blob/4d2d4d47a18d93f3e0a91a241a6fdb545bbf7dc2/sshconnect.c#L1106-L1118

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,3 +1,8 @@
+AllCops:
+  Exclude:
+    - 'tryout/**/*'
+  NewCops: enable
+
 inherit_from: .rubocop_todo.yml
 
 Style/DoubleNegation:
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,6 +1,8 @@
 AllCops:
   Exclude:
     - 'tryout/**/*'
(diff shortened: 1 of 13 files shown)


fwininger:
Looks good to me.

Issue #819, opened 2021-07-01
Support ruby's frozen_string_literal option


j1wilmot:
### Problem

Running applications dependent on `net-ssh` with `RUBYOPT=--enable-frozen-string-literal` results in `FrozenError`s.

This behavior can be demonstrated by running the test suite with the `frozen-string-literal` option enabled:
```sh
RUBYOPT=--enable-frozen-string-literal bundle exec rake test
```

### Solution

Add magic comments to alert ruby that string mutations are expected.

`# frozen_string_literal: false` can be added on specific files that mutate strings, or across the board on all files. This is similar to what was done in ruby itself in https://github.com/ruby/ruby/commit/3e92b635fb5422207b7bbdc924e292e51e21f040 and will prevent the `FrozenError` from being raised when a string is mutated while ruby has the frozen_string_literal option enabled.

Happy to open a PR if there is interest in this approach. Thanks for the great gem!

fwininger:
@j1wilmot we happy if you can help on this with a PR.

fwininger:
close with #835

PR #835, opened 2021-08-07
Support frozen_string_literals

mfazekas:
Fixes: #720 

Closes: #821, #819


fwininger:
Better solution than add the comment on the top. Thanks :)

PR #846, opened 2022-01-11
Check for timeouts while reading headers during server version negotiation

bschmeck:
Currently, if we connect to a non-SSH server, we'll potentially block forever waiting to receive an SSH version string.  We'll read whatever data the server sends into `@header` but that data could be anything, e.g., the banner message from an FTP server.  If the server never sends a line that starts with `SSH-`, we'll wait around until the server breaks the connection.

This PR moves the `IO::select` timeout inside of the reading loop, so that we check for an IO timeout after each newline from the server.  We could still hang if the server never sends a final newline, but calling `IO.select` after reading each byte felt excessive.  Alternatively, we could use a combination of `IO#read_nonblock` and `IO::select` to implement a version of `IO#readpartial` with timeouts.

This might also merit a separate `read_timeout` option, since this implementation doesn't enforce a hard limit on the time it takes to connect.  (E.g. if the timeout is 10 and the server sleeps for 8 seconds between sending each line, we'll wind up taking far longer than 10 seconds to connect, or eventually timeout if the server never sends an `SSH-` line.)

I haven't attempted to add a test yet, because I wasn't sure if this is a good direction to go.  I'm happy to write a test for this implementation, though.  Just let me know.

### DIFF
--- a/lib/net/ssh/transport/server_version.rb
+++ b/lib/net/ssh/transport/server_version.rb
@@ -45,16 +45,23 @@ module Net
           socket.write "#{PROTO_VERSION}\r\n"
           socket.flush
 
-          loop do
-            raise Net::SSH::ConnectionTimeout, "timeout during server version negotiating" if timeout && !IO.select([socket], nil, nil, timeout)
+          raise Net::SSH::ConnectionTimeout, "timeout during server version negotiating" if timeout && !IO.select([socket], nil, nil, timeout)
 
+          loop do
             @version = String.new
             loop do
               begin
-                b = socket.readpartial(1)
+                b = socket.read_nonblock(1)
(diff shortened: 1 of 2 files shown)


mfazekas:
@bschmeck for first glance this looks ok to me.

bschmeck:
Thanks.  It looks like #784 might fix the same problem, I can check in the morning.  That PR uses `IO#read_nonblock` and `IO::select` to effectively add timeouts to `IO#readpartial`, which feels like a better solution than my approach here.

bschmeck:
#784 was raising timeout errors as expected when I tested it locally this morning.  I'll close this PR in favor of that one.

Thanks for all your work!

PR #849, opened 2022-02-21
openssl: DSA: don't hardcode expected signature size

terceiro:
The default value of the Q parameters for DSA keys changed in Ruby
OpenSSL 3.0.0, and that causes DSA signatures to be longer by default.
This change might have been accidental, and this may be reverted; see
https://github.com/ruby/openssl/issues/483

This changes the check for the signature length to not be against a
hardcoded expected lenght, but against the expected length as calculated
from the Q parameter.


fwininger:
Thanks for the PR :)

PR #850, opened 2022-03-02
Support `~` in the path to the SSH agent's unix socket

boblail:
`.ssh/config` supports paths in this style but `UNIXSocket.open` does not expand paths

Notably, environment variables like `$USER` are not supported by `ssh` for the `IdentityAgent` field so

```
Host *
  IdentityAgent "~/Library/Application Support/xyz/agent.sock"
```
works; but
```
Host *
  IdentityAgent "$USER/Library/Application Support/xyz/agent.sock"
```
does not

### DIFF
--- a/lib/net/ssh/authentication/agent.rb
+++ b/lib/net/ssh/authentication/agent.rb
@@ -85,9 +85,9 @@ module Net
             if agent_socket_factory
               agent_socket_factory.call
             elsif identity_agent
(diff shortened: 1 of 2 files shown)


boblail:
@mfazekas, I based this off of [v6.1.0](https://github.com/net-ssh/net-ssh/tree/v6.1.0). Could this also be merged to a 6.1 branch — and shipped as 6.1.1?

PR #853, opened 2022-03-25
ed25519 pubkey in ssh format, including comment (when present)

marvinthepa:
Allows getting the public key in ssh-format (including the comment), from an original public key or private key

```
  public_key.to_ssh

  private_key.public_key.to_ssh
```

I was not sure about the method name to_ssh, but it fits the (apparently broken) to_pem method.

I would have liked to also implement to_ssh for the private key, but the format is more complicated and I do not know all the details - I also do not need it right now :-). Let me know if you want both key types to have it.

Also it seems a bit harder to also parse the comment when parsing the public key (as it is not part of the binary blob).

Let me know if you think this needs a test or anything else, will be happy to improve this PR.
Also simplified some outdated comments, hope that's okay.

### DIFF
--- a/lib/net/ssh/authentication/ed25519.rb
+++ b/lib/net/ssh/authentication/ed25519.rb
@@ -106,9 +106,11 @@ module Net
           include Net::SSH::Authentication::PubKeyFingerprint
 
           attr_reader :verify_key
+          attr_accessor :comment
 
-          def initialize(data)
+          def initialize(data, comment = "")
             @verify_key = ::Ed25519::VerifyKey.new(data)
+            @comment = comment
           end
 
           def self.read_keyblob(buffer)
@@ -135,6 +137,10 @@ module Net
             # TODO this is not pem
             ssh_type + Base64.encode64(@verify_key.to_bytes)
           end
+
+          def to_ssh
+            "#{ssh_type} #{Base64.strict_encode64(to_blob)} #{comment}"
+          end
         end
 
         class PrivKey
@@ -144,12 +150,12 @@ module Net
           MEND = "-----END OPENSSH PRIVATE KEY-----\n"
           MAGIC = "openssh-key-v1"
 
-          attr_reader :sign_key
+          attr_reader :sign_key, :comment
 
           def initialize(buffer)
             pk = buffer.read_string
             sk = buffer.read_string
-            _comment = buffer.read_string
+            @comment = buffer.read_string
 
             @pk = pk
             @sign_key = SigningKeyFromFile.new(pk, sk)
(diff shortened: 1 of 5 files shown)


lib/net/ssh/authentication/ed25519.rb
```diff
@@ -135,6 +137,10 @@ def to_pem
             # TODO this is not pem
             ssh_type + Base64.encode64(@verify_key.to_bytes)
           end
+
+          def to_ssh
```
marvinthepa, line 141: could also override `to_s`.

This is the behavior of subclasses of OpenSSL::PKey::PKey, which are the other classes returned from `Net::SSH::KeyFactory::load_*_key`.

mfazekas (commented):
Looks good to me, maybe you can add a option to return without comment with `to_ssh` then `to_pem` could call just `to_ssh`, we seems to use `to_pem` to compare to serialised keys. 

Also some simple tests would be needed.

marvinthepa:
> maybe you can add a option to return without comment with `to_ssh` then `to_pem` could call just `to_ssh`, we seems to use `to_pem` to compare to serialised keys.

I think even without the comment `to_pem` does something else than to_ssh (missing one " " between the key type and the binary blob for starters - but it seems the binary is also different). Do you think there will be no problem with keys serialized with the old `to_pem` if you change the format?

> Also some simple tests would be needed.

I will add some.

test/authentication/test_ed25519.rb
```diff
@@ -24,6 +24,8 @@ def test_no_pwd_key
         signed = priv_key.ssh_do_sign(shared_secret)
         self.assert_equal(true, pub_key.ssh_do_verify(signed, shared_secret))
         self.assert_equal(priv_key.public_key.fingerprint, pub_key.fingerprint)
+        self.assert_equal(priv_key.comment, "vagrant@vagrant-ubuntu-trusty-64")
```
marvinthepa, line 27: IMHO `self` is redundant here. But I added it as the surrounding code was using it.

Should I fix it in the whole file?

test/authentication/test_ed25519.rb
```diff
@@ -24,6 +24,8 @@ def test_no_pwd_key
...
+        self.assert_equal(priv_key.public_key.comment, "vagrant@vagrant-ubuntu-trusty-64")
```
marvinthepa, line 28: AFAIK the order is wrong here, usually it is

  expected, actual

not the other way around.

I added it in a consistent way to the surrounding code, do you want me to fix it for the whole file? Should that be an extra commit?

mfazekas:
> > maybe you can add a option to return without comment with `to_ssh` then `to_pem` could call just `to_ssh`, we seems to use `to_pem` to compare to serialised keys.
> 
> I think even without the comment `to_pem` does something else than to_ssh (missing one " " between the key type and the binary blob for starters - but it seems the binary is also different). Do you think there will be no problem with keys serialized with the old `to_pem` if you change the format?

I think `to_pem` is only used to compare if to keys is identical, so changing the format should not cause issue.

PR #859, opened 2022-04-12
Fix link in Readme

fwininger:


fwininger:
#858 introduce a typo

PR #860, opened 2022-04-12
Use the last ubuntu LTS

fwininger:
Remplace ubuntu 18.04 with ubuntu 20.04

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,7 @@ name: CI
 on: [push, pull_request]
 jobs:
   test:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


VitSkakun:
please note that latest Ubuntu LTS is 22.04 not 20.04

just for the record in Ubuntu versions XX.YY 
XX means year
YY - month
and LTS releases are issued on April every second year
so 20.04 is LTS release issued on some April day in 2020

.github/workflows/ci.yml
```diff
@@ -2,7 +2,7 @@ name: CI
 on: [push, pull_request]
 jobs:
   test:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
```
VitSkakun, line 5: please note that latest Ubuntu LTS version is 22.04 issued ~1week ago

Issue #874, opened 2022-06-30
With OpenSSL 3, net-ssh 7 cannot use private keys provided through start options of ssh config file


guss77:
### Expected behavior

With net-ssh 7.0.1 and openssl 3.0.2, the following code should be able to connect to the remote SSH server:

```
Net::SSH.start(host, user, { keys: [ 'key.pem' ] })
```

### Actual behavior

`Net::SSH.start` fails to connect and throws `Net::SSH::AuthenticationFailed` exception with the message `Authentication failed for user username@hostname`. When setting the option `verbose: Logger::DEBUG`, I get these messages:

```
E, [2022-06-30T12:00:10.511537 #1037434] ERROR -- net.ssh.authentication.key_manager[76c]: could not load private key file `key.pem': OpenSSL::PKey::PKeyError (rsa#set_key= is incompatible with OpenSSL 3.0)
E, [2022-06-30T12:00:10.511572 #1037434] ERROR -- net.ssh.authentication.session[730]: all authorization methods failed (tried none, publickey)
```

### System configuration

- net-ssh version 7.0.1
- Ruby version 3.0.2p107

### Example App

```ruby
#!/usr/bin/env ruby

gem 'net-ssh', '=7.0.1'

require 'net/ssh'

host = 'geek.co.il'
user = 'ubuntu'
opts = { keys: [ '~/.ssh/private/aws-vhost.pem' ], verbose: Logger::DEBUG }
Net::SSH.start(host, user, opts) do |ssh|
	puts ssh.exec!('cat /etc/machine-id')
end
```

guss77:
Here's a video showing the behavior of the test app vs. openssh client

https://user-images.githubusercontent.com/381782/176640617-56763566-e633-47ef-894c-545f11aa99d8.mp4

schopin-pro:
Following up on @dentarg 's comment on the other issue, to implement the RSA private key support in the same way as it's been done for the public key, you'll need to create an ASN.1 representation of the key. The best resource I could find that describes that representation (or rather, those representations, as it depends whether we're using PKCS#1 or PKCS#8) is this blog article:

https://www.thedigitalcatonline.com/blog/2018/04/25/rsa-keys/

nikita015:
I'm also getting 
```
[...]

PR #875, opened 2022-07-15
Don't use the deprecated set_XXX methods on RSA keys.

bschmeck:
This addresses the issue with using RSA keys with OpenSSL v3, as detailed in #874.  This code is largely taken from https://github.com/nov/json-jwt/pull/102, just renaming a few variables.

Running `rake test` in a container built from `Dockerfile.openssl3` succeeds on both master and this branch.  I'd be happy to write a test for this change, but would need some pointers on where it should live, if it should test `read_private_keyblob` directly or a higher level method, etc.


kkurian:
@schopin-pro Any thoughts on this PR? Does it address the issue in #874? Does it raise concerns?

lib/net/ssh/buffer.rb
```diff
@@ -262,27 +261,28 @@ def read_private_keyblob(type)
...
-            key.p = p
-            key.q = q
+          # Public key
+          data_sequence = OpenSSL::ASN1::Sequence([
+                                                    OpenSSL::ASN1::Integer(n),
+                                                    OpenSSL::ASN1::Integer(e),
+                                                  ])
+
+          if d && p && q && dmp1 && dmq1 && iqmp
+            data_sequence = OpenSSL::ASN1::Sequence([
```
schopin-pro, line 271: Since we're in the `read_private_keyblob` method, wouldn't it make more sense to just abort/raise an exception rather than going through the motions of actually creating the pubkey?

schopin-pro:
It seems reasonable :)

lib/net/ssh/buffer.rb
bschmeck: I'm definitely out of my depth, but why would we want to raise an error here?

The existing code constructs a `PKey` but uses the newly deprecated `set_key`, `set_factors` and `set_crt_params` methods when doing so.  I was under the impression that building a key directly from the ASN.1 gets us the same result, but has the benefit of not using the deprecated methods.

fdr:
FYI: we were just adjusting our code base to be OpenSSL 3 compatible and updated net-ssh, only to find this problem, which prevents us from using the Heroku-22 binary stack. Very much a live issue.

will:
I also ran into this recently when trying to upgrade to openss3. It gives me `OpenSSL::PKey::PKeyError rsa#set_key= is incompatible with OpenSSL 3.0`, so I had to go back to the older version of openssl

isaacbowen:
In my testing over here, this issue only comes up with non-PEM key data. On openssl 3 and net-ssh 7.0.1, I have zero issue if I first convert to PEM format, before loading in the key.

Does that line up with anyone else's experience?

guss77:
> In my testing over here, this issue only comes up with non-PEM key data. On openssl 3 and net-ssh 7.0.1, I have zero issue if I first convert to PEM format, before loading in the key.

Same for me - after converting private key files to PEM format, the test code in issue #874 completes correctly using net-ssh 7.0.1.

Alternatively, after applying the patch in this PR (after rebasing on net-ssh `master`), the aforementioned test code completes successfully **without** converting the private key file - using the original OpenSSH private key format.

I would really appreciate it if this PR can be merged and a new public net-ssh release can be done. Currently, I can use this patch on my system but getting it to work in a bundler application is not trivial and every new install requires too much manual wrangling.

If @bschmeck is unavailable to update the PR (it basically needs rebasing) I can adopt it - please comment if you prefer I adopt this.

bschmeck:
I'm still here :)

I just merged `master` into my branch and pushed.

cjilbert504:
Just want to chime in and say that I recently hit this issue and this PR fixes it for me. Thanks so much!

excid3:
Any chance on getting this reviewed / merged @mfazekas? We ran into this a couple times recently.

fwininger (approved):

fwininger:
looks good @excid3 @cjilbert504. 
Thanks @bschmeck @schopin-pro

andygeorge (commented):
LGTM

excid3:
Thanks @fwininger 👍

mbarany:
Thanks for the fix! Could we please get a release going? 🙏

mbarany:
@mfazekas yes that works. thank you!

guss77:
I can also confirm working for me, without additional workarounds.

mfazekas:
7.1.0 final was released - https://rubygems.org/gems/net-ssh/versions/7.1.0

PR #883, opened 2022-10-11
Fix certificate signature valid check

sshock:
This fixes the `Certificate#signature_valid?` check, which was failing on "rsa-sha2-512" signatures because it was always using SHA1.


sshock:
It would be great if we could get this merged in, as it is affecting my company's ability to upgrade our Linux machines due to newer OS using openssl 3 which requires net-ssh 7.

Issue #890, opened 2023-01-04
connection closed by remote host (bad packet being sent by net-ssh)


znewsham:
### Expected behavior
Should be able to connect to an SSH host that accepts connections from the ssh command line tool

### Actual behavior
connections created from net-ssh fail after the type 50 packet is sent (debug trace below)

The tech support for the host we are connecting to claims: 
> Your client is trying to make a connection but during the securing part of the SSH tunnel, your side is sending a packet(doing something weird) and that packet is not reassembled on our side.

> Is your library/client RFC Compliant?

Ordinarily we'd assume they were being silly - but the fact that we can connect from the command line and not from net-ssh suggests the problem may be with net-ssh.

```
D, [2023-01-04T17:42:52.231753 #610] DEBUG -- net.ssh.transport.session[258]: establishing connection to safetrans.wellsfargo.com:22
D, [2023-01-04T17:42:52.313043 #610] DEBUG -- net.ssh.transport.session[258]: connection established
I, [2023-01-04T17:42:52.313138 #610]  INFO -- net.ssh.transport.server_version[26c]: negotiating protocol version
D, [2023-01-04T17:42:52.313147 #610] DEBUG -- net.ssh.transport.server_version[26c]: local is `SSH-2.0-Ruby/Net::SSH_7.0.1 x86_64-linux'
D, [2023-01-04T17:42:52.343562 #610] DEBUG -- net.ssh.transport.server_version[26c]: remote is `SSH-2.0-Axway.Gateway'
I, [2023-01-04T17:42:52.343725 #610]  INFO -- net.ssh.transport.algorithms[280]: sending KEXINIT
D, [2023-01-04T17:42:52.343838 #610] DEBUG -- socket[294]: queueing packet nr 0 type 20 len 1436
D, [2023-01-04T17:42:52.343917 #610] DEBUG -- socket[294]: sent 1440 bytes
D, [2023-01-04T17:42:52.371464 #610] DEBUG -- socket[294]: read 280 bytes
D, [2023-01-04T17:42:52.371505 #610] DEBUG -- socket[294]: received packet nr 0 type 20 len 276
I, [2023-01-04T17:42:52.371520 #610]  INFO -- net.ssh.transport.algorithms[280]: got KEXINIT from server
I, [2023-01-04T17:42:52.371541 #610]  INFO -- net.ssh.transport.algorithms[280]: negotiating algorithms
[...]

PR #891, opened 2023-01-05
Accept pubkey_algorithms option when starting a new connection

RemcodM:
a45f54fe1de434605af0b7195dd9a91bccd2cec5 introduced the option to perform RSA client authentication using newer algorithm types, e.g. SHA2.

For me personally, this has broken `net-ssh` with a few servers. We use `net-ssh` in combination with `net-sftp` to communicate with a few obscure SSH implementations that we cannot control. It seems that some of these servers sent a disconnect when trying to authenticate with these newer algorithms, which are tried by default before the old SHA1 algorithm. This might also be the cause of #890, although the servers that we try to communicate with do actually still sent a disconnect packet (resulting in a `Net::SSH::Disconnect` exception), while that does not seem to be the case in #890.

a45f54fe1de434605af0b7195dd9a91bccd2cec5 also introduces the option `pubkey_algorithms` to a `Session` which allows the user of the `Session` to control which algorithms are used for client authentication. However, this option is not available to `Net::SSH#start`, which means that passing in that option results in:
```
ArgumentError: invalid option(s): pubkey_algorithms
```

This PR changes `Net::SSH#start` such that `pubkey_algorithms` is allowed as an option. This allows us to use `sha-rsa` as algorithm for some of the more obscure SSH implementations that we are dealing with, such that we can successfully connect again.

### DIFF
--- a/lib/net/ssh.rb
+++ b/lib/net/ssh.rb
@@ -73,7 +73,7 @@ module Net
       max_win_size send_env set_env use_agent number_of_password_prompts
       append_all_supported_algorithms non_interactive password_prompt
       agent_socket_factory minimum_dh_bits verify_host_key
-      fingerprint_hash check_host_ip
+      fingerprint_hash check_host_ip pubkey_algorithms
     ]
 
     # The standard means of starting a new SSH connection. When used with a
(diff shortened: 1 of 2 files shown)


znewsham:
While this does solve errors when connecting to specific hosts - I suspect it will cause new errors for hosts that don't use that protocol. 

I believe the root cause of the issue in #890 is that the negotiated `host_key` is not used and instead each option in `pubkey_algorithms` is tried in turn. For loosely strict servers, this probably works (method 1 is attempted, fails, method 2 attempts, passes), but for servers that strictly implement the protocol, it seems they disconnect when they receive an auth method they don't support.

I believe https://github.com/net-ssh/net-ssh/pull/892 is closer to the correct solution, but don't know enough about the protocol (and naming conventions within net-ssh) to be sure

saarons:
@znewsham I'm also affected by this error. I think the server you had the original issue on is technically non-compliant but also it's probably impossible for them to turn off that behavior:

From the RFC about `SSH_MSG_USERAUTH_REQUEST` messages:

```
While there is usually little point for clients to send requests that
   the server does not list as acceptable, sending such requests is not
   an error, and the server SHOULD simply reject requests that it does
   not recognize.
```

znewsham:
@saarons we figured the same - we ended up with a much more complicated fix (we connect to ~100 different servers) it's definitely not a complete solution, but it seems to work: https://github.com/znewsham/net-ssh/pull/1

saarons:
@znewsham That actually looks like a great fix, I might run that against our list as well and see what happens.

mfazekas (approved):

PR #892, opened 2023-01-05
Force the use of the negotiated host_key algorithm

znewsham:
another attempt at fixing issue #890 - I believe the root cause of the issue is that we don't use the negotiated host_key when submitting the auth request, instead iterating through the client supported pub key algs. 

It would appear that strict servers disconnect when this happens.


znewsham:
This isn't right :(

PR #893, opened 2023-01-17
Add none support to ProxyJump

elmobp:
As per OpenSSH man docs

```
ProxyJump
             Specifies one or more jump proxies as either
             [user@]host[:port] or an ssh URI.  Multiple proxies may be
             separated by comma characters and will be visited
             sequentially.  Setting this option will cause ssh(1) to
             connect to the target host by first making a ssh(1)
             connection to the specified ProxyJump host and then
             establishing a TCP forwarding to the ultimate target from
             there.  Setting the host to none disables this option
             entirely.
```

This along with the tests should obey the above

### DIFF
--- a/lib/net/ssh/config.rb
+++ b/lib/net/ssh/config.rb
@@ -153,8 +153,10 @@ module Net
 
             # ProxyCommand and ProxyJump override each other so they need to be tracked togeather
             %w[proxyjump proxycommand].each do |proxy_key|
-              if (proxy_value = settings.delete(proxy_key))
-                settings['proxy'] ||= [proxy_key, proxy_value]
+              unless proxy_key == 'proxyjump' and proxy_value == 'none'
+                if (proxy_value = settings.delete(proxy_key))
+                  settings['proxy'] ||= [proxy_key, proxy_value]
+                end
               end
             end
           end
--- a/lib/net/ssh/config.rb
+++ b/lib/net/ssh/config.rb
@@ -154,6 +154,7 @@ module Net
             # ProxyCommand and ProxyJump override each other so they need to be tracked togeather
             %w[proxyjump proxycommand].each do |proxy_key|
               if (proxy_value = settings.delete(proxy_key))
+                # When ProxyJump is set to none explicity disable the jumphost
                 unless proxy_key == 'proxyjump' and proxy_value == 'none'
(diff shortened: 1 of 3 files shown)


lib/net/ssh/config.rb
```diff
@@ -154,7 +154,10 @@ def load(path, host, settings = {}, base_dir = nil)
             # ProxyCommand and ProxyJump override each other so they need to be tracked togeather
             %w[proxyjump proxycommand].each do |proxy_key|
               if (proxy_value = settings.delete(proxy_key))
-                settings['proxy'] ||= [proxy_key, proxy_value]
+                # When ProxyJump is set to none explicity disable the jumphost
```
mfazekas, line 157: Shouldn't we clear settings['proxy'] if proxy jump is none. I mean if one config has a proxy configured but a more specific sets it none then we should clear it, shouldn't we?

elmobp: Sorry I totally missed this! Let me fix that up

lib/net/ssh/config.rb
```diff
@@ -155,6 +155,10 @@ def load(path, host, settings = {}, base_dir = nil)
             %w[proxyjump proxycommand].each do |proxy_key|
               if (proxy_value = settings.delete(proxy_key))
                 settings['proxy'] ||= [proxy_key, proxy_value]
+                # When ProxyJump is set to none explicity remove the proxy settings
```
mfazekas, line 158:
```suggestion
```

Pls remove the comment as it adds no information, just repeats the code

elmobp:
@mfazekas I totally forgot about this! anything you need from me?

lib/net/ssh/config.rb
```diff
@@ -155,6 +155,9 @@ def load(path, host, settings = {}, base_dir = nil)
             %w[proxyjump proxycommand].each do |proxy_key|
               if (proxy_value = settings.delete(proxy_key))
                 settings['proxy'] ||= [proxy_key, proxy_value]
+                if proxy_key == 'proxyjump' and proxy_value == 'none'
+                  settings.delete('proxy')
+                end
```
jeremy, line 160: Need to leave the setting in place so later ProxyJump / ProxyCommand directives don't override it.

PR #900, opened 2023-02-15
Timeout when waiting for server version string

bschmeck:
When connecting to a server that doesn't return a version string (usually a non-SSH server), connections will currently hang indefinitely.  This change introduces a timeout during version negotiation to avoid that situation.

Previously #784 and #846.


bschmeck:
The hang can be recreated by running a server with netcat and connecting to it via `::start`.

In one terminal: `nc -l -p 4444` 
In a separate terminal: `Net::SSH.start("localhost", "user", password: "password", port: 4444)`

fwininger (approved):

mfazekas:
> The hang can be recreated by running a server with netcat and connecting to it via `::start`.
> 
> In one terminal: `nc -l -p 4444` In a separate terminal: `Net::SSH.start("localhost", "user", password: "password", port: 4444)`

But if you use the timeout argument then we do get an exception.

Net::SSH.start("localhost", "user", password: "password", port: 4444,timeout: 10)
*** Net::SSH::ConnectionTimeout Exception: timeout during server version negotiating

See also: #247

mfazekas (changes requested):
I'd like this addressed first before we proceed with this PR 

https://github.com/net-ssh/net-ssh/pull/900#issuecomment-1445970108

bschmeck:
@mfazekas Sorry, my test setup was overly simplified.  This PR addresses the situation where we can connect to a remote server and it replies with data but not a server version.  If you pass text to the netcat server (e.g. `echo "this is not an SSH server" | nc -l -p 4444`), then the master branch hangs while this branch raises an error.

I can work on adding a test like the one in #902 if you think that will be helpful.

mfazekas:
> @mfazekas Sorry, my test setup was overly simplified. This PR addresses the situation where we can connect to a remote server and it replies with data but not a server version. If you pass text to the netcat server (e.g. `echo "this is not an SSH server" | nc -l -p 4444`), then the master branch hangs while this branch raises an error.
> 
> I can work on adding a test like the one in #902 if you think that will be helpful.

@bschmeck that would be great, FWIW we have some integration tests using TCPServer, so that could be used.

https://github.com/net-ssh/net-ssh/blob/da52b108fded76af7bcf896340cd3e3862e74d66/test/integration/test_forward.rb#L49-L66

bschmeck:
@mfazekas Took me longer than I expected to get around to this, but I just pushed an integration test for this change.  It mimics the test added in #902 but sends data back from the netcat server to get us into the read loop during version negotiation.  The test fails on master and passes on this branch, but there seems to be an issue in CI due to how Github Actions is setting up test environments.

mfazekas:
@bschmeck thanks much, I finally got around fixing our CI in GitHub actions.

bschmeck:
:wave: I've let this PR sit for too long, but I believe that I've addressed your concerns.  Please let me know if there's something else I should do before this is ready to be merged.  Thanks!

mfazekas:
@bschmeck thanks much sorry will try to look into it in the following days

bschmeck:
Thanks for all your work maintaining the project.  I'm in no rush, just trying to clean up some branches I have laying around.

PR #901, opened 2023-02-22
Terminate proxy command when session is closing

rtib:
Under some circumstances, when using SSH ProxyCommand, the process running the proxy might not terminated on session shutdown. This patch is terminating the proxy process when the session was set up using ProxyCommand.

### DIFF
--- a/lib/net/ssh/transport/session.rb
+++ b/lib/net/ssh/transport/session.rb
@@ -126,6 +126,7 @@ module Net
         # Cleans up (see PacketStream#cleanup) and closes the underlying socket.
         def close
           socket.cleanup
+          Process.kill('TERM', socket.pid) if options[:proxy].is_a?(Net::SSH::Proxy::Command)
           socket.close
         end
 


mfazekas:
@rtib thanks much for the PR, can you describe the issue in more details, can you provide a sample so we can reproduce it?

rtib:
@mfazekas can we talk about that on a private channel?

lib/net/ssh/transport/session.rb
```diff
@@ -126,6 +126,7 @@ def closed?
         # Cleans up (see PacketStream#cleanup) and closes the underlying socket.
         def close
           socket.cleanup
+          Process.kill('TERM', socket.pid) if options[:proxy].is_a?(Net::SSH::Proxy::Command)
```
mfazekas, line 129: I don't really like this solution. options[:proxy], if anything this should be moved to Proxy::Command. the returned io close method could easily do this kill. But I'd really like to see the concrete use case first

mfazekas (changes requested):

inopinatus:
I can confirm I've seen this problem when using EC2 Instance Connect, which creates a websocket-based tunnel for ProxyCommand and is fine with regular openssh but hangs on close with Net::SSH.

However, this PR doesn't fully solve the problem. It only kills off the local proxy process and doesn't properly disconnect the remote server.

This is specifically bad for things like Instance Connect. Being over a websocket, the remote proxy doesn't know the local proxy has died - it looks more like it just went quiet. I guess it's designed to handle a transient loss of connectivity. The remote server is never disconnected and unfortunately we can still see the user session, they're never cleanly logged out, until they time out an hour or so later.

Even this is enough to trip it over:

```Net::SSH.start('i-123456789', 'deploy') { |ssh| puts ssh.exec!("id") }```

in the presence of

```ProxyCommand aws ec2-instance-connect open-tunnel --instance-id %h```

Is it possible that Net:SSH isn't handling disconnect in exactly the same fashion as the openssh client? For a proxy designed like EC2 Instance Connect, that could be significant.

A possible workaround is to set ClientAliveInterval on the server to a low number (e.g. 5) and ClientAliveCountMax to a higher number (e.g. 10). Interestingly this not only causes the connection to get shut down on both ends, it happens on the very first keepalive after the socket.close - the server never gets to ClientAliveCountMax. However, the additional traffic has side-effects, and the local proxy command dies with a horrible looking exception error, so that's another imperfect fix.

rtib:
My assumption is, that if the proxy is writing something to e.g. stderr, that channel doesn't close automatically when the server is closing its connection.

However, I like the approach in #947 .

Issue #477, opened 2016-12-29
Support chacha20-poly1305@openssh.com cipher


tisba:
### Expected behavior

I'm trying to harden SSH on servers I manage. My goal would be able to use

```
KexAlgorithms curve25519-sha256@libssh.org
HostKeyAlgorithms ssh-ed25519
Ciphers chacha20-poly1305@openssh.com
```

Which works very well with `ssh`. For deployments using `net-ssh` via Capistrano, I have to add more (arguably less secure) ciphers.

Cipher `chacha20-poly1305@openssh.com` should be supported.

### Actual behavior

(tested using capistrano)

```
Net::SSH::Exception: could not settle on encryption_client algorithm
```

### System configuration

- net-ssh 4.0
- capistrano 3.7.1
- Ruby 2.3.1/2.3.3/2.4

mfazekas:
Thanks for the report, unfortunately it's not likely to be implemented in the near future unless there is a PR, it sounds it either require some native code or it might be implemented on top of as rbnacl 4.0 seems have it.

See
https://github.com/cryptosphere/rbnacl/blob/master/lib/rbnacl/aead/chacha20poly1305_ietf.rb

See
https://github.com/openssh/openssh-portable/blob/master/PROTOCOL.chacha20poly1305
https://github.com/openssh/openssh-portable/blob/master/chacha.c
https://github.com/openssh/openssh-portable/blob/master/cipher-chachapoly.c

fwininger:
@tisba : I don't have a PR for chacha20-poly1305, but I have one for "KexAlgorithms curve25519-sha256@libssh.org".

fwininger:
relative to #705

HLFH:
Any updates?

```
cap aborted!
SSHKit::Runner::ExecuteError: Exception while executing as deploy@example.com: could not settle on encryption_client algorithm
Server encryption_client preferences: chacha20-poly1305@openssh.com
Client encryption_client preferences: aes256-ctr,aes192-ctr,aes128-ctr


Caused by:
Net::SSH::Exception: could not settle on encryption_client algorithm
Server encryption_client preferences: chacha20-poly1305@openssh.com
Client encryption_client preferences: aes256-ctr,aes192-ctr,aes128-ctr
```

fwininger:
#845

mfazekas:
It's now implemented in 7.2.0.beta1. Please not it's only active if you have RbNaCl gem installed. Pls test and report back.

[...]

PR #908, opened 2023-05-15
chacha20-poly1305 with RbNaCl poly1305

mfazekas:
Fixes: #477 

Implementation of `chacha20-poly1305` chipher.

Since openssl has no separate poly implementation it useless RbNaCl for chacha20 and poly1305 implementation. So this feature is only enabled if you have rbnacl gem installed in your system.

TODO:
- [x] uses poly1305 from rbnacl. We can treat that as an optional dependency and only add chacha20-poly1305 if that's installed
- [x] tests
- [x] refactor internals since chach20-poly1305 has mac embedded I've added a bunch of == "chacha20-poly1305" that needs to be simplified

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,6 +77,12 @@ jobs:
         env:
           NET_SSH_RUN_INTEGRATION_TESTS: 1
           CI: 1
+      - name: Run tests (without rbnacl)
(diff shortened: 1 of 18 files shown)


Fryguy:
FWIW this is being associated with CVE-2023-48795

PR #912, opened 2023-06-13
Allow IdentityAgent as option to Net::SSH.start

### LINKED ISSUE #911: Net::SSH.start does not work with configs with IdentityAgent options
### Expected behavior

SSH connects properly

### Actual behavior

```
/***/lib/ruby/gems/3.1.0/gems/net-ssh-7.2.0.beta1/lib/net/ssh.rb:225:in `start': invalid option(s): identity_agent (ArgumentError)
```

### System configuration

- net-ssh version
  - 7.2.0.beta1
- Ruby version
  - 3.1.4

### Example App

Please provide an example script that reproduces the problem. This will save maintainers time so they can spend it fixing your issues instead of trying to build a reproduction case from sparse instructions.

You can use this as stating point:

```ruby
gem 'net-ssh', '= 7.2.0.beta1'
rrequire "net/ssh"

@host = "localhost"
@user = "user"
@identity_agent = "/path/to/auth.sock"

Net::SSH.start(@host, @user, {
  :identity_agent => @identity_agent,
}) do |ssh|
end
```

tosuke:
Fixes #911

### DIFF
--- a/lib/net/ssh.rb
+++ b/lib/net/ssh.rb
@@ -192,6 +192,7 @@ module Net
     #   Defaults to %w(~/.ssh/known_hosts ~/.ssh/known_hosts2).
     # * :use_agent => Set false to disable the use of ssh-agent. Defaults to
     #   true
+    # * :identity_agent => the path to the ssh-agent's UNIX socket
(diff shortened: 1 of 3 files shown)


mfazekas (approved):
👍

Issue #653, opened 2019-01-09
Jump host disrecards `keys` option


akostadinov:
### Expected behavior

When using [jump host](https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/proxy/jump.rb), the proxy command should be generated with inserted `-i` options for each of the `keys` for the connection.

### Actual behavior

`-i` options are not inserted and connection fails

### System configuration

- net-ssh version
Net::SSH::Version::STRING "5.0.2"

- Ruby version
"2.5.3"

### Example App

```ruby
gem 'net-ssh', '= 5.0.2'
require 'net/ssh'
require 'net/ssh/proxy/jump'
puts Net::SSH::Version::CURRENT

@host = 'localhost'
@user = ENV['USER']
jump_host = Net::SSH::Proxy::Jump.new("user@some-host")
Net::SSH.start(@host, @user, keys: ["key1", "key2"], proxy: jump_host) do |ssh|
  puts ssh.exec!('echo "hello"')
end
```

P.S. Isn't there some way to proxy a new Net::SSH connection through an existing one? Perhaps if one has ability to create a port forwarding in existing connection, then that port can be used as target for new connection.

mfazekas:
@akostadinov on the actual issue i'm not sure if we always use the same keys for jump host as for target hosts. I don't think OpenSSH has this behaviour, does it?


BTW are you aware of the https://github.com/net-ssh/net-ssh-gateway project, it might be what you looking for.

mfazekas:
A way to add additional arguments to Proxy::Jump might be the simplest solution.

akostadinov:
Why specify additional options, when `keys` are already specified?

Issue #617, opened 2018-07-18
Jump Proxy ignores verify_host_key and other SSH configuration


adammw:
### Expected behavior

When using a jump proxy, the jump proxy applies the same configuration as the rest of net-ssh. 

### Actual behavior

When using a jump proxy, net-ssh shells out to the system's ssh command, and doesn't pass along configuration such as turning off host key verification. 

### System configuration

- net-ssh version: 5.0.2
- Ruby version: 2.5.1p57

### Example App

```ruby
gem 'net-ssh', '= 5.0.2'
require 'net/ssh'
puts Net::SSH::Version::CURRENT

@host = 'localhost'
@user = ENV['USER']
proxy = Net::SSH::Proxy::Jump.new('user@proxy')
Net::SSH.start(@host, @user, proxy: proxy, verify_host_key: :never) do |ssh|
  puts ssh.exec!('echo "hello"')
end
```

mfazekas:
@adammw i'm not sure about this request. I'd say you might want to verify proxy but not target host or vica versa, as per 653 a way to pass additional options/arguments to Proxy::Jump might be enough.

Issue #596, opened 2018-04-06
Using a 2nd key configured on a Proxy::Jump?


jletizia:
I was curious if the following was possible (I have yet to figure out the right combination of runtime options to make this work):

Host A authenticates to Proxy::Jump host using Key A, the Proxy::Jump host has Key B stored on it, which is then used to authenticate against the final endpoint of Host B.

Is this possible? All of the documentation that I've read (utilizing the keys option, etc) seem to only support specifying a list of keys to get to the Proxy::Jump host in the first place, but not specifying a 2nd key to use once there. Is there any sort of ssh config file fun that I could do instead to achieve the same thing?

jletizia:
Basically the behavior I'm trying to reproduce is:

`ssh -t <JUMP-USER>@<JUMP-IP> -p <JUMPPORT> ssh <ENDHOST>@<ENDHOST-IP>`

where it will connect to <JUMP-IP> using my local key, and then connect to <ENDHOST-IP> using <JUMP-USER>'s local key stored on <JUMP-IP>

mfazekas:
Adding an way to add more arguments to `Jump::Proxy` per #653 might solve this too.

PR #915, opened 2023-06-30
Support SSH options and keys in Jump Proxy

excid3:
This adds support for SSH options on the command line with the -o flag. You can pass them in as options when you instantiate the Jump class.

```ruby
proxy = Net::SSH::Proxy::Jump.new('user@proxy', options: ["StrictHostKeyChecking=no"])
```

Keys and key_data from the SSH connection are also passed to the ssh proxy. The key_data strings are written to disk as Tempfiles and cleaned up when the Jump object is garbage collected to ensure they exist when the ssh command is run.

This addresses a couple of issues like #653, #596, #617 and can be expanded to support other flags.


lib/net/ssh/proxy/jump.rb
```diff
@@ -10,7 +11,7 @@ module Proxy
       #
       #   require 'net/ssh/proxy/jump'
       #
-      #   proxy = Net::SSH::Proxy::Jump.new('user@proxy')
+      #   proxy = Net::SSH::Proxy::Jump.new('user@proxy', options: ["StrictHostKeyChecking=no"])
```
excid3, line 14: Another option here would be making this more generic like:

```ruby
proxy = Net::SSH::Proxy::Jump.new('user@proxy', flags: ["-o StrictHostKeyChecking=no"])
```

Then users could pass in any SSH flags which would be more flexible.

Issue #920, opened 2023-08-09
Pubkey authentication results in SSH::Disconnect from server in Net-SSH 7.2


nathansamson:
### Expected behavior

Net-SSH does not connect anymore in net-ssh 7.2 with a certain server (this might very well be a server issue, but it worked in net-ssh 6.2).

I do not control the server or can access their logs. I can help providing debug info from my side if possible...

### Actual behavior

Server disconnects

~~~
D, [2023-08-09T18:29:29.351860 #68872] DEBUG -- net.ssh.transport.session[97cc]: establishing connection to redacted:8022
D, [2023-08-09T18:29:29.361517 #68872] DEBUG -- net.ssh.transport.session[97cc]: connection established
I, [2023-08-09T18:29:29.361602 #68872]  INFO -- net.ssh.transport.server_version[97e0]: negotiating protocol version
D, [2023-08-09T18:29:29.361617 #68872] DEBUG -- net.ssh.transport.server_version[97e0]: local is `SSH-2.0-Ruby/Net::SSH_7.2.0 x86_64-linux'
D, [2023-08-09T18:29:29.371049 #68872] DEBUG -- net.ssh.transport.server_version[97e0]: remote is `SSH-2.0-JSCAPE'
I, [2023-08-09T18:29:29.371301 #68872]  INFO -- net.ssh.transport.algorithms[97f4]: sending KEXINIT
D, [2023-08-09T18:29:29.371560 #68872] DEBUG -- socket[9808]: queueing packet nr 0 type 20 len 860
D, [2023-08-09T18:29:29.371647 #68872] DEBUG -- socket[9808]: sent 864 bytes
D, [2023-08-09T18:29:29.374154 #68872] DEBUG -- socket[9808]: read 360 bytes
D, [2023-08-09T18:29:29.374238 #68872] DEBUG -- socket[9808]: received packet nr 0 type 20 len 356
I, [2023-08-09T18:29:29.374290 #68872]  INFO -- net.ssh.transport.algorithms[97f4]: got KEXINIT from server
I, [2023-08-09T18:29:29.374360 #68872]  INFO -- net.ssh.transport.algorithms[97f4]: negotiating algorithms
D, [2023-08-09T18:29:29.374553 #68872] DEBUG -- net.ssh.transport.algorithms[97f4]: negotiated:
* kex: diffie-hellman-group-exchange-sha256
* host_key: ssh-rsa
* encryption_server: aes256-ctr
[...]

PR #923, opened 2023-08-31
Add support for PubkeyAcceptedAlgorithms option

zzambers:
Support for configuring `pubkey_algorithms` in configuration file using [PubkeyAcceptedAlgorithms](https://man.openbsd.org/ssh_config#PubkeyAcceptedAlgorithms). Also `pubkey_algorithms` is no longer limited to RSA.

Can be useful to workaround cases, where ssh server (wrongly) disconnects on unknown algorithm, without need to modify program code. (See e.g. https://github.com/net-ssh/net-ssh/issues/890 https://github.com/net-ssh/net-ssh/pull/891 https://github.com/net-ssh/net-ssh/issues/920)

Configuration examples:
```
Host host1
  PubkeyAcceptedAlgorithms -rsa-sha2-256

Host host2
  PubkeyAcceptedAlgorithms ssh-rsa,ssh-dss
```

### DIFF
--- a/lib/net/ssh/authentication/methods/abstract.rb
+++ b/lib/net/ssh/authentication/methods/abstract.rb
@@ -20,9 +20,7 @@ module Net
           # this.
           attr_reader :key_manager
 
-          # So far only affects algorithms used for rsa keys, but can be
-          # extended to other keys, e.g after reading of
-          # PubkeyAcceptedAlgorithms option from ssh_config file is implemented.
+          # public key algorithms to be used
           attr_reader :pubkey_algorithms
 
           # Instantiates a new authentication method.
@@ -31,11 +29,7 @@ module Net
             @key_manager = options[:key_manager]
             @options = options
             @prompt = options[:password_prompt]
-            @pubkey_algorithms = options[:pubkey_algorithms] \
-              || %w[rsa-sha2-256-cert-v01@openssh.com
-                    ssh-rsa-cert-v01@openssh.com
-                    rsa-sha2-256
-                    ssh-rsa]
+            @pubkey_algorithms = session.transport.pubkey_algorithms
             self.logger = session.logger
           end
 
--- a/lib/net/ssh.rb
+++ b/lib/net/ssh.rb
@@ -170,11 +170,10 @@ module Net
     # * :properties => a hash of key/value pairs to add to the new connection's
     #   properties (see Net::SSH::Connection::Session#properties)
     # * :proxy => a proxy instance (see Proxy) to use when connecting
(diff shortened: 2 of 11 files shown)


zzambers:
There is not that much happening in `algorithms.rb` by this PR, as we only have list of public key algorithms configured on client side so far. However code can be later further extended to support `server-sig-algs` extension, to get list of public key algorithms supported by server, if available. Final list can than be created with algorithms listed by both parties. (similar as for other algorithms)

lib/net/ssh/authentication/methods/abstract.rb
```diff
@@ -31,11 +29,7 @@ def initialize(session, options = {})
             @key_manager = options[:key_manager]
             @options = options
             @prompt = options[:password_prompt]
-            @pubkey_algorithms = options[:pubkey_algorithms] \
-              || %w[rsa-sha2-256-cert-v01@openssh.com
```
mfazekas, line 35: Our default pubkeys_alogirthm will be much extended, won't it? I mean in case confg doesn't define it. 


https://github.com/net-ssh/net-ssh/pull/923/files#diff-d3f3cd24513081ec6736e7516607d8de54cb329fe6319a2c3fabe4edd67b4393R51-R65

zzambers: Yy, It is because `pubkey_algorithms` is now checked for all keys (not just RSA). (See [change in lib/net/ssh/authentication/methods/publickey.rb](https://github.com/net-ssh/net-ssh/pull/923/files#diff-f39e09d93c6f4a8374520565aa7ca0bbca6ed2c614305bc1d8997b9263612461L126))

I have included all public key algorithms, I have found in net-ssh sources, in defatults, so that default behavior remains the same (all of them still allowed), as `pubkey_algorithms` now also affects keys other than RSA.  I also kept order of RSA algorithms, relative to each other (to keep behavior for RSA keys).

zzambers: RSA keys are somewhat special, because they are the only ones, where there is not 1:1 relationship between keys and algorithms. There are multiple possible algorithms for one key (since [RFC 8332](https://datatracker.ietf.org/doc/html/rfc8332)).

Variable `pubkey_algorithms` was originally internal, used to just control order of RSA algorithms (with possibility to be extended for other algorithms later, i.e. `PubkeyAcceptedAlgorithms`). As `pubkey_algorithms` was made public by: https://github.com/net-ssh/net-ssh/pull/891 , I would like to fully implement it.

PR #926, opened 2023-09-13
Load keycert directly from string

mkasztelnik:
My application uses short-lived SSH credentials generated on the fly. The user SSH key and a certificate are stored in the DB. Currently to establish the connection I need to use the following code:

```ruby
Tempfile.create do |f|
  f << user.ssh_certificate
  f.flush
  puts Net::SSH.start("my.host", "username", 
                      key_data: [user.ssh_key], keycerts: [f.path], 
                      keys_only: true) { |ssh| ssh.exec!("hostname") }
end
```

This PR adds the missing element - the possibility to pass the key certificate as a string. As a result, the code above can be simplified to:

```ruby
Net::SSH.start("my.host", "username", 
               key_data: [user.ssh_key], keycert_data: [user.ssh_certificate], 
               keys_only: true) { |ssh| ssh.exec!("hostname") }
```

### DIFF
--- a/lib/net/ssh.rb
+++ b/lib/net/ssh.rb
@@ -66,7 +66,7 @@ module Net
       auth_methods bind_address compression compression_level config
       encryption forward_agent hmac host_key identity_agent remote_user
       keepalive keepalive_interval keepalive_maxcount kex keys key_data
-      keycerts languages logger paranoid password port proxy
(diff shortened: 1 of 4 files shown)


mfazekas (approved):

mkasztelnik:
@mfazekas is there anything more I should add/improve to get it merged?

PR #928, opened 2023-10-03
Ruby 3.3.0-preview2: Fix error when reading from a closed stream

intrip:
Fixes "TypeError, no implicit conversion of nil into String" when reading from a closed stream. This error is related to a change in the behavior of Socket#recv which was introduced in Ruby 3.3.0.

See https://bugs.ruby-lang.org/issues/19899?issue_count=1&issue_position=1

### DIFF
--- a/lib/net/ssh/buffered_io.rb
+++ b/lib/net/ssh/buffered_io.rb
@@ -61,7 +61,7 @@ module Net
       # if no data was available to be read.
       def fill(n = 8192)
         input.consume!
-        data = recv(n)
+        data = recv(n) || ""
         debug { "read #{data.length} bytes" }
         input.append(data)
         return data.length


intrip:
I've checked the other parts in the codebase where `recv` is used and haven't found any other issues because there's no direct String operation done on the result.

mfazekas (approved):

mattbrictson:
@mfazekas is it possible to get this fix included in a new release of net-ssh to rubygems soon? I think the general release of Ruby 3.3 is less than 2 weeks away.

(For context, capistrano breaks on Ruby 3.3.0-rc1 https://github.com/capistrano/capistrano/issues/2146 due to this bug)

mfazekas:
@mattbrictson I've released https://rubygems.org/gems/net-ssh/versions/7.2.1.rc1 pls test, I'll release 7.2.1 during next week.

mattbrictson:
@mfazekas I've confirmed that ssh-kit and capistrano work fine with 7.2.1.rc1 👍

mfazekas:
I've just released 7.2.1

PR #929, opened 2023-11-13
Raise specified error when key-files are not readable

topstone:
It is useful to raise specified error  when key-files are not readable. (Easy to deconstruct the problem.)

I'm not sure it is the best way to improve.


topstone:
Here is the minimum file to raise error. File "ghost_key" does not exist.

```
require 'net/ssh'

Net::SSH.start('my.server.com', 'myname', keys: ["./ghost_key"]) do |ssh|
end
```

mfazekas:
@topstone I'm not 100% about this. It's a breaking change as before it ignored non existing key files, now we're throwing...

Not sure what openssh behaviour is in such cases

PR #932, opened 2023-12-26
Remove base64 runtime dependency

mattbrictson:
Starting in Ruby 3.3.0, `require "base64"` triggers a warning:

> base64 was loaded from the standard library, but will no longer be part of the default gems since Ruby 3.4.0. Add base64 to your Gemfile or gemspec.

Rather than introduce a new dependency, this commit removes the base64 requirement in favor of using `String#pack` and `String#unpack1` directly. (The base64 gem is just a thin wrapper around these methods.)

I left base64 as a development dependency so that the existing tests could continue to use `Base64`, to confirm there are no regressions due to my changes.

The approach in this PR is similar to that used in several other gems, including:

- https://github.com/lostisland/faraday/pull/1541
- https://github.com/rack/rack/pull/2110
- https://github.com/rubocop/rubocop/pull/12313
- https://github.com/patsplat/plist/pull/63

### DIFF
--- a/lib/net/ssh/authentication/ed25519.rb
+++ b/lib/net/ssh/authentication/ed25519.rb
@@ -46,7 +46,7 @@ module Net
             raise ArgumentError.new("Expected #{MEND} at end of private key") unless datafull.end_with?(MEND)
 
             datab64 = datafull[MBEGIN.size...-MEND.size]
-            data = Base64.decode64(datab64)
+            data = datab64.unpack1("m")
             raise ArgumentError.new("Expected #{MAGIC} at start of decoded private key") unless data.start_with?(MAGIC)
 
             buffer = Net::SSH::Buffer.new(data[MAGIC.size + 1..-1])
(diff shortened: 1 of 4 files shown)


mfazekas (approved):

mfazekas:
@mattbrictson thanks much, this look good!

mfazekas:
@mattbrictson I've released this as https://rubygems.org/gems/net-ssh/versions/7.2.2.rc1

mattbrictson:
@mfazekas testing against 7.2.2.rc1, and unfortunately I still see the warning:

> net-ssh-7.2.2.rc1/lib/net/ssh/known_hosts.rb:3: warning: base64 was loaded from the standard library, but will no longer be part of the default gems since Ruby 3.4.0. Add base64 to your Gemfile or gemspec. Also contact author of net-ssh-7.2.2.rc1 to add base64 into its gemspec.

It looks like this PR was incomplete: I forgot to remove the `require "base64"` statements. I'll open another PR.

PR #942, opened 2024-03-23
Key manager can load public key as IdentityFile

kpumuk:
This PR allows setting public keys as `IdentityFile` in SSH config.

> If an SSH client supports setting public keys as `IdentityFile`, you can use that to [match hosts to a specific key in 1Password](https://developer.1password.com/docs/ssh/agent/advanced/#match-key-with-host).

This enables deployments with Capistrano or Kamal using SSH keys stored in 1Password when also pinning the public key to a specific host.

```
Host deploy
  IdentitiesOnly yes
  IdentityFile ~/.ssh/kamal.pub
  IdentityAgent ~/.1password/agent.sock
```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,5 +2,8 @@
 .git/
 .gitignore
 
+# Vagrant
+test/integration/.vagrant
+
 docker-compose.yml
 README.md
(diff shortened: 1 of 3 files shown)


tylerhunt:
Working for me to allow Kamal deployment. Hope to see this merged.

PR #947, opened 2024-05-11
Tell any ProxyCommand process to hang up.

inopinatus:
If there's a pid in the IO, send it a SIGHUP on cleanup. This solves a problem when using EC2 Instance Connect Endpoints, which have peculiar semantics and don't handle shutdown across the tunnel the way a regular socket might, leading to zombie sessions on the server and net-ssh hanging on close.

May resolve other similar issues with unconventional proxy configurations. Found by comparison with OpenSSH, which also sends a HUP, so this is behavioural parity¹.

Obviously a variant on #901, which suggests TERM but that proved unreliable for EICE. This change is in a different place with a different signal and I wanted the specific case on record.

------------

_edit to add_: there is an `at_exit` version of this but I don't believe libraries have any business installing `at_exit` handlers². For folks wanting that, I suggest

```
Net::SSH::Proxy::Command.prepend Module.new {
  def open(*)
    super.tap do |io|
      at_exit { Process.kill('HUP', io.pid) rescue nil }
    end
  end
}
```

in your program, Capfile or what have you, and now the command process will be signaled even if the parent exits unexpectedly.

------------

¹ https://github.com/openssh/openssh-portable/blob/1c0d81357921f8d3bab06841df649edac515ae5b/sshconnect.c#L276

² The clearest reason not to do this by default is how it hurts long-lived process. Foremost is memory leakage, since that `at_exit` handler is a closure binding the io object, so both are now forever on the heap, and this builds up with every connection. It also has a bonus nasty surprise failure mode in long-lived processes when pids get reused. Long-running programs would instead need an approach for tracking pids that fits their circumstances.

### DIFF
--- a/lib/net/ssh/transport/packet_stream.rb
+++ b/lib/net/ssh/transport/packet_stream.rb
@@ -187,6 +187,7 @@ module Net
         def cleanup
           client.cleanup
           server.cleanup
+          Process.kill('HUP', pid) if pid
         end
 
         # If the IO object requires a rekey operation (as indicated by either its
--- a/test/transport/test_packet_stream.rb
+++ b/test/transport/test_packet_stream.rb
@@ -69,6 +69,7 @@ module Transport
     def test_cleanup_should_delegate_cleanup_to_client_and_server_states
       stream.client.expects(:cleanup)
       stream.server.expects(:cleanup)
+      stream.expects(:pid)
       stream.cleanup
     end
 
@@ -1102,6 +1103,7 @@ module Transport
             assert packet[:always_display]
             assert_equal "debugging", packet[:message]
             assert_equal "", packet[:language]
+            stream.stubs(:pid).returns(nil)
             stream.cleanup
           end
 
@@ -1120,6 +1122,7 @@ module Transport
             stream.client.set cipher: cipher, hmac: hmac, compression: compress
             stream.enqueue_packet(ssh_packet)
             assert_equal PACKETS[cipher_name][hmac_name][compress], stream.write_buffer
+            stream.stubs(:pid).returns(nil)
             stream.cleanup
           end
         end


inopinatus:
Discovered when using Capistrano, so I'll document my temporary workaround, appending to the Capfile:

```
# Fix EC2 Instance Connect hangs
Net::SSH::Transport::PacketStream.prepend Module.new {
  def cleanup
    super
    Process.kill('HUP', pid) if pid
  end
}
```

dptsec:
Independent confirmation of this bug, plus the reproduction that was asked for over in #901.

We hit it deploying with Kamal 2.12 (net-ssh 7.3.3) through an EC2 Instance Connect Endpoint ProxyCommand (`aws ec2-instance-connect open-tunnel ...`). The deploy completes successfully, then the process hangs forever at exit. Sampling the hung process shows exactly where:

```
rb_ec_exec_end_proc          (at_exit cleanup)
  → rb_io_close_m            (IO#close on the proxy pipe)
    → rb_waitpid → __wait4   (waiting for a child that never exits)
```

`Proxy::Command` returns an `IO.popen` pipe as the socket, and Ruby's `IO#close` on a popen'd IO waits for the child to exit — which a tunnel command that doesn't quit on stdin EOF never does. OpenSSH is immune because it signals its ProxyCommand child on session teardown, which is the parity this PR restores.

Minimal reproduction, no AWS required:

```ruby
require 'net/ssh/proxy/command'

proxy = Net::SSH::Proxy::Command.new('echo ready && sleep 60')
io = proxy.open('example.com', 22)
io.close  # blocks until the sleep exits; with a real tunnel process, forever
```

There's also a compounding effect worth noting: every interrupted run strands the tunnel process (reparented to PID 1), and for EICE the stranded tunnels count against the endpoint's concurrency limit — after a few, new connections hang for minutes during `open`, which makes the whole thing look intermittent and much harder to diagnose.

Two small data points that may help this land:

- Signal choice: current awscli (2.34) exits promptly on TERM as well, but HUP matches OpenSSH's behavior on teardown, so the choice here seems right.
- It may be worth rescuing `Errno::ESRCH` around the `Process.kill`, for cleanup paths where the command has already been reaped (double cleanup, or applications running their own child reapers) — a dead proxy shouldn't raise mid-teardown.

We wrote behavioral regression tests for this (real popen'd command that outlives EOF; assert `close` returns within a timeout and the process is actually gone; double-close idempotency) — they fail with `Timeout::Error` on unpatched master and pass with a fix in place. Happy to contribute them to this branch; they're on https://github.com/dptsec/net-ssh/tree/proxy-command-kill-child-on-close (a variant of this fix at the `Proxy::Command` level) and are portable to this PR's approach as-is.

PR #948, opened 2024-05-21
Support for `exec` in host matching and token substitution in other ssh_config keywords

davisonja:
The goal of the PR is to support the ssh_config file options like:
```
Match host vm.example.com exec "/usr/local/bin/cloudflared access ssh-gen --hostname %h"
    HostName vm.example.com
    ProxyCommand /usr/local/bin/cloudflared access ssh --hostname %h
    IdentityFile ~/.cloudflared/vm.example.com-cf_key
    CertificateFile ~/.cloudflared/vm.example.com-cf_key-cert.pub
```
which relate to [Cloudflare Zero Trust](https://developers.cloudflare.com/cloudflare-one/identity/users/short-lived-certificates/#7-connect-as-a-user) ssh connections.

Through a slightly convoluted series of steps I'm keen to see support for token expanision and/in `exec` clauses in the ssh config files - which leads to this PR!

 * Processing of tokens has been pulled into a single function for re-use
 * tokens are substituted based on a pre-built hash of `token: value`
 * the hash is built either from ssh_config style parameters, or net ssh style parameters; as this PR only includes `remote user` and `host`, the difference looks to reduce to just string vs symbol index, but they're still different so it's been left as two functions
 * tokens are expanded in other keywords beyond `exec`, some of which come from the use case that inspired the PR, and others because they supported token expansion
 * evaluation of the `match` condition has a changed regex

The last of these is probably the biggest change. When using `exec` in a match condition the command to execute is specified after a ` `, but supports whitespace in the command by quoting the value. The existing regex didn't support quoted values, and thus split the `exec` command 'in the middle'.
The new regex correctly parsed all the existing examples in the test suite, and others (from my use case), however, changes in regex always benefit from more eyes and checks!


lib/net/ssh/proxy/command.rb
```diff
@@ -38,25 +38,16 @@ def initialize(command_line_template)
...
-              else
-                raise ArgumentError, "remote user name not available"
-              end
-            when '%'
-              '%'
-            else
-              raise ArgumentError, "unknown key: #{$1}"
+          net_ssh_settings = { host: host, port: port }
+          if connection_options
+            net_ssh_settings.merge!(connection_options)
```
mfazekas, line 43: Is this `merge!` necessary?

mfazekas:
@davisonja thanks much for the PR! 👍 

Can you please  add a short summary of the what is the PR addressing in description?

PR #950, opened 2024-08-23
Refactor Object#=~ as it is deprecated since Ruby 2.6 and unsupported since Ruby 3.2

srbarrios:
As you can read [here](https://rubyreferences.github.io/rubychanges/3.2.html#removals), if we want that net-ssh still work in Ruby >= 3.2, we need to change `=~` by `match` method.

That's a quick PR addressing only that particular breaking change.
`test` Rake task pass, but you might want additional pre-condition that the object calling the method `match`, is not actually a `nil` object, otherwise it will raise an error. Let me know what do you think about it, but I could only see one place where that really was necessary.

### DIFF
--- a/lib/net/ssh.rb
+++ b/lib/net/ssh.rb
@@ -92,7 +92,7 @@ module Net
     #
     #         ch.on_data do |ch, data|
     #           print data
-    #           if data =~ /sudo password: /
+    #           if data.match(/sudo password: /)
     #             ch.send_data("password\n")
     #           end
     #         end
--- a/lib/net/ssh/buffered_io.rb
+++ b/lib/net/ssh/buffered_io.rb
@@ -173,7 +173,7 @@ module Net
           debug { "connection was reset => shallowing exception:#{e}" }
           return 0
         rescue IOError => e
-          if e.message =~ /closed/ then
+          if e.message.match(/closed/) then
             debug { "connection was reset => shallowing exception:#{e}" }
             return 0
           else
@@ -189,7 +189,7 @@ module Net
           debug { "connection was reset => shallowing exception:#{e}" }
           return 0
         rescue IOError => e
-          if e.message =~ /closed/ then
+          if e.message.match(/closed/) then
             debug { "connection was reset => shallowing exception:#{e}" }
             return 0
           else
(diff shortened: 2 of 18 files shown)


dentarg:
Is this solving a real life problem you are experiencing? I assume many net-ssh users are using net-ssh with Ruby 3.2 and 3.3.

`=~` hasn't been removed?

I think the change you have made could be risky

```ruby
$ RUBY_OPTS=-w ruby -e 'p RUBY_VERSION; nil =~ /foo/'
"3.3.4"

$ RUBY_OPTS=-w ruby -e 'p RUBY_VERSION; nil.match?(/foo/)'
"3.3.4"
-e:1:in `<main>': undefined method `match?' for nil (NoMethodError)

p RUBY_VERSION; nil.match?(/foo/)
                   ^^^^^^^
```

dentarg:
I think the actual problem, where this happens, should be found and addressed

```ruby
$ RUBY_OPTS=-w ruby -e 'p RUBY_VERSION; Object.new =~ /foo/'
"3.3.4"
-e:1:in `<main>': undefined method `=~' for an instance of Object (NoMethodError)
```

Just switching to `#match?` will still fail equally hard?

```ruby
$ RUBY_OPTS=-w ruby -e 'p RUBY_VERSION; Object.new.match?(/foo/)'
"3.3.4"
-e:1:in `<main>': undefined method `match?' for an instance of Object (NoMethodError)

p RUBY_VERSION; Object.new.match?(/foo/)
                          ^^^^^^^
```

mfazekas:
@srbarrios  thanks much the PR

>if we want that net-ssh still work in Ruby >= 3.2, we need to change =~ by match method.
I think this statement is misleading.
 
`=~` is working fine in Ruby >= 3.2, we expect the left side to be `string` in most of code changed. Sorry  this PR sounds a bit like cargo culting. If you see specific issue with `=~` that's another issue, and we're happy to address that, but as I currently see it just adds changes, which could either improve or make things worse. Thanks for the understanding.

srbarrios:
Sorry guys, I misunderstood the issue I was facing, all good.
Thanks for the fast replies.

PR #953, opened 2024-09-17
Stop remote forwarding from logging to stdout

djmb:
Send debug statements to the logger instead of stdout. Though maybe we could remove them altogether?


mfazekas:
@djmb 👍   that was my bad, some debugging for integration tests.  I think debug is ok.

https://github.com/net-ssh/net-ssh/pull/806/files

PR #958, opened 2024-10-17
Support frozen strings

npezza93:


lib/net/ssh/test/extensions.rb
```diff
@@ -57,7 +57,7 @@ def idle!
             return false unless script.next(:first)
 
             if script.next(:first).remote?
-              self.string << script.next.to_s
+              self.string = "#{self.string.dup}#{script.next}"
```
Lukom: no need in `dup` here, just:

```ruby
self.string = "#{self.string}#{script.next}"
```

morgoth:
@mfazekas FYI: Looks like this is good to go now.

mfazekas (changes requested):

morgoth:
@mfazekas This now looks like can be merged

mfazekas (approved):

mfazekas:
@morgoth will try to release on the weekend pls ping me if I forgot

morgoth:
@npezza93 Yes, that was it.

@mfazekas I tested with net-ssh 7.3.1.rc1 and net-scp from master and now kamal works fine with frozen string enabled.
Would be great to have both gems released.

Thanks.

PR #963, opened 2024-11-20
Fix host and ip assignment in HostKeyError

atharvaksh:
If the host is a URL then host_as_string will be "fileshare.com,123.123.123.1" which can split into the host as the first and IP as the second part.

If the host is an IP address then host_as_string will be "123.123.123.1" so the host and IP both are assigned the same value


atharvaksh:
@mfazekas please take a look at this, thanks!

PR #964, opened 2024-12-10
Fix parsing of the algorithms from ssh_config

sztupy:
Based on `ssh_config`'s documentation, section `Ciphers` parsing of the algorithm configs work in the following way :

https://www.man7.org/linux/man-pages/man5/ssh_config.5.html

> Specifies the ciphers allowed and their order of preference.  Multiple ciphers must be comma-separated.  If the specified list begins with a ‘+’ character, then the specified
> ciphers will be appended to the default set instead of replacing them.  If the specified list begins with a ‘-’ character, then the specified ciphers (including wildcards) will
> be removed from the default set instead of replacing them.  If the specified list begins with a ‘^’ character, then the specified ciphers will be placed at the head of the
> default set.

This means only the very first character in the list is used to determine operation (append, prepend, remove), afterwards the rest of the config items will remain as-is, without any prefix needed

This change would modify how the config is parsed from the previous version, which expected each item to have it's own prefix. This was likely due to the wrong wording in older versions of the config's documentation that was fixed a while ago: https://github.com/openssh/openssh-portable/commit/4f9d75fbafde83d428e291516f8ce98e6b3a7c4b It also adds support for the `^` prepend operator added here: https://github.com/openssh/openssh-portable/commit/91a2135f32acdd6378476c5bae475a6e7811a6a2

More specifically this fixes an issue with recent version of MacOS (Sequoia 15 and above) where the default ssh config is the following:

```
Ciphers ^aes-128-gcm@openssh.com,aes256-gcm@openssh.conf
```

Current version of the code would drop the first item (as it doesn't understand the prepend operator), then add the second. If the first character would be a `+` then however the first item would have been added, with the second silently ignored, although correct operation would have been to add both to the end of the list.

### DIFF
--- a/lib/net/ssh/transport/algorithms.rb
+++ b/lib/net/ssh/transport/algorithms.rb
@@ -302,22 +302,28 @@ module Net
           list = []
           option = Array(option).compact.uniq
 
-          if option.first && option.first.start_with?('+', '-')
+          if option.first && option.first.start_with?('+', '-', '^')
             list = supported.dup
 
-            appends = option.select { |opt| opt.start_with?('+') }.map { |opt| opt[1..-1] }
-            deletions = option.select { |opt| opt.start_with?('-') }.map { |opt| opt[1..-1] }
+            modifier = option.first[0]
+            options_without_modifier = [option.first[1..-1]] + option[1..-1]
 
-            list.concat(appends)
-
-            deletions.each do |opt|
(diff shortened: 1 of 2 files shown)


scottgeary:
Would love this to get more attention.

sztupy:
Not sure if it's possible to ping the maintainer @mfazekas

PR #966, opened 2025-03-08
Add logger gem dependency to gemspec

javierav:
This PR remove warning about incoming changes in ruby 3.5.0:

```
/Users/.../.asdf/installs/ruby/3.4.2/lib/ruby/gems/3.4.0/gems/net-ssh-7.3.0/lib/net/ssh.rb:5:
warning: logger was loaded from the standard library, but will no longer be part of the default
gems starting from Ruby 3.5.0.
You can add logger to your Gemfile or gemspec to silence this warning.
```


cleeland:
+1 encountered this same issue and came here to see if it's already fixed.

eglitobias:
+1 Would be good if we can merge this.

Issue #965, opened 2024-12-13
Test failures with zlib-ng


voxik:
On recent Fedora, we observe test failures such as:

~~~
$ ruby -Ilib:test test/transport/test_packet_stream.rb -n test_enqueue_packet_with_aes192_cbc_and_hmac_md5_96_and_standard_compression
Run options: -n test_enqueue_packet_with_aes192_cbc_and_hmac_md5_96_and_standard_compression --seed 61672

# Running:

F

Finished in 0.003381s, 295.7541 runs/s, 295.7541 assertions/s.

  1) Failure:
Transport::TestPacketStream#test_enqueue_packet_with_aes192_cbc_and_hmac_md5_96_and_standard_compression [test/transport/test_packet_stream.rb:1138]:
--- expected
+++ actual
@@ -1,3 +1,3 @@
 # encoding: ASCII-8BIT
 #    valid: true
-"se\xE7\x98\x16\xC9\x8A\xA8yH\xA1\xC2n\xF4:\xBE\xB8M=H1\xCF\x92^\xF2\x9FD\x95N\xEC:\xE3\x85M\x06[\xCB$U/yZ\xD8\x9D\x1A\xC7\xD0D-\xE5\b\x15\xA2#\xDEP8\xE3\xF1\xB3"
+"\xDC14b\xB2\xE754\xDB6\x0F3\xDF\xB7,eYb\xFB\x87t\x99\xCE\xFB\xCFo\xB4~\xDDA_\x02Z\xD1vx\xDC\xA7\xF5\x16j\x18J\x14}\x06z#\x9A\x1CG\x8F\x1AuN\xA6\xFE\x05\xD1\xDB"


1 runs, 1 assertions, 1 failures, 0 errors, 0 skips
~~~

And we believe this is because Fedora is using [zlib-ng](https://github.com/zlib-ng/zlib-ng) in place of zlib. Is there a chance to make the test suite more robust to handle different zlib implementations? BTW on some HW, there might be use also different optimizations, providing different results even between platforms.

mfazekas:
@voxik it would be great, but current tests has hardwired results atm

junaruga:
[...]

PR #976, opened 2025-09-11
CI: Add zlib-ng case

junaruga:
@mfazekas, this PR is related to https://github.com/net-ssh/net-ssh/issues/965 and https://github.com/net-ssh/net-ssh/issues/975#issuecomment-3271384975, and adds the [zlib-ng](https://github.com/zlib-ng/zlib-ng) case in CI. The assertion failing in the zlib-ng case is skipped for now. Because I couldn't find an easy way to pass the failing assertion without hardcoding the expected results, and also it was hard to add the hardcoded strings for each cipher_method_name and hmac_method_name case.

Add zlib-ng case in CI using Fedora Linux that uses zlib-ng instead of zlib.
https://src.fedoraproject.org/rpms/zlib-ng

The assertion failing in the zlib-ng case is skipped for now.

What do you think?

### DIFF
--- a/.github/workflows/ci-with-docker.yml
+++ b/.github/workflows/ci-with-docker.yml
@@ -42,3 +42,17 @@ jobs:
 
     - name: Run ruby 3.0 with openssl3
       run: docker run --rm netssh_openssl3
+
+  test_zlib_ng:
+    name: Run test suite with docker and zlib-ng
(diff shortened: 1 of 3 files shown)


junaruga:
Thanks for merging this PR!

PR #984, opened 2025-12-27
Add logger & fiddle as runtime dependencies

bastelfreak:
starting with Ruby 4, logger isn't bundled into Ruby anymore and it needs to be explicitly declared as dependency.

### DIFF
--- a/net-ssh.gemspec
+++ b/net-ssh.gemspec
@@ -30,6 +30,8 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
+  spec.add_runtime_dependency 'logger', '< 2'
+
   unless ENV['NET_SSH_NO_ED25519']
     spec.add_development_dependency("bcrypt_pbkdf", "~> 1.0") unless RUBY_PLATFORM == "java"
     spec.add_development_dependency("ed25519", "~> 1.2")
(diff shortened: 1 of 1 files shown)


bastelfreak:
@mfazekas any chance you could merge & release this change to add Ruby 4 support?

bastelfreak:
@mfazekas hey, any chance you could release this?

net-ssh.gemspec
```diff
@@ -30,6 +30,9 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
+  spec.add_runtime_dependency 'fiddle', '< 2'
```
mfazekas: Not sure about fiddle it's a needed only on windows. And ruby gems has no good way of expressing that.

mfazekas (changes requested):

net-ssh.gemspec
bastelfreak: You can do windows specific releases and then scope certain dependencies to a platform. nokogiri for example does windows specific releases. But I never digged around to figure out how this works :(

bastelfreak: If you prefer, I can remove fiddle for now?

mfazekas: > If you prefer, I can remove fiddle for now?

Yeah let's do that now.

nokigiri ha a c extension, so of course it has a windows specific release. We don't have C extension, but we could still do a windows specific release, but it requires release process tweaks, etc. Windows users on ruby 4 need to declare the fiddle in their own gem file, it's not optimal but works for now

bastelfreak: perfect, updated it

bastelfreak: @mfazekas can you please take a look and merge it?

PR #990, opened 2026-03-03
Handle inline comments in ssh config

saraid:
I'm kinda confused how it was working before: I could swear I had inline comments in my config and it worked fine. But I don't have a record of that (could have been a lot of versions ago), and it isn't working now.

The issue is simply that a config like
```
Host foo
  SettingA 1    # this is set because of X
  SettingB xyz  # this is set because of Y
```
results in `config[:setting_b]` being `xyz  # this is set because of Y` instead of `xyz`.

I think that I've set everything up correctly, but first time contribution, so lmk if I need to do anything else.

### DIFF
--- a/lib/net/ssh/config.rb
+++ b/lib/net/ssh/config.rb
@@ -399,7 +399,7 @@ module Net
         end
 
         def uncomment(string)
-          string[...string =~ /\s+#.*?$/]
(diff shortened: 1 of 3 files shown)


mfazekas (changes requested):
Thanks for the PR!

PR #997, opened 2026-05-09
Stop mutating ENV['HOME'] at require time

### LINKED ISSUE #996: Feature Request / Bug: Avoid global mutation of ENV['HOME'] on require for in-memory key operations
Currently, net-ssh mutates global environment state the moment the gem is required. Specifically, in `lib/net/ssh.rb`, the library executes the following logic to enforce a fallback for `$HOME`:

Ruby
```
ENV['HOME'] ||= ENV['HOMEPATH'] ? "#{ENV['HOMEDRIVE']}#{ENV['HOMEPATH']}" : Dir.pwd
```
While I understand this was implemented to prevent `File.expand_path` from crashing when looking up local SSH configs or identity files, modifying global state (ENV) at require-time creates unintended side effects.

We run stripped-down environments where `ENV['HOME']` is intentionally unset. We use net-ssh strictly for in-memory key parsing and manipulation—specifically using `Net::SSH::KeyFactory.load_data_public_key` to read raw string data.

Because we aren't touching the filesystem, net-ssh has no actual need to resolve a home directory. However, simply requiring the gem silently populates `ENV['HOME']` with `Dir.pwd` for the entire Ruby process, which inadvertently affects other dependencies that rely on the absence of `$HOME` to trigger their own logic (we're changing this too). 

Can net-ssh avoid eagerly mutating `ENV['HOME']` on load?

Perhaps instead of globally patching the environment variable, the home directory fallback could be evaluated lazily. For example, the library could implement an internal helper method (e.g., `Net::SSH.home_directory`) that performs this fallback logic only when a filesystem operation actually needs to resolve a tilde ~ path.

mfazekas:
Fixes #996

Remove the eager `ENV['HOME']` mutation from `lib/net/ssh.rb` line 3. Ruby 2.6+ (net-ssh's minimum) resolves `~` paths via the OS passwd database when HOME is unset, making this unnecessary. The few `File.expand_path` call sites that handle `~` default paths now rescue `ArgumentError` to degrade gracefully.

### DIFF
--- a/lib/net/ssh.rb
+++ b/lib/net/ssh.rb
@@ -1,7 +1,3 @@
-# Make sure HOME is set, regardless of OS, so that File.expand_path works
-# as expected with tilde characters.
-ENV['HOME'] ||= ENV['HOMEPATH'] ? "#{ENV['HOMEDRIVE']}#{ENV['HOMEPATH']}" : Dir.pwd
-
 require 'logger'
 require 'etc'
 require 'shellwords'
--- a/lib/net/ssh/authentication/key_manager.rb
+++ b/lib/net/ssh/authentication/key_manager.rb
@@ -72,11 +72,15 @@ module Net
         def add(key_file)
           key_files.push(File.expand_path(key_file)).uniq!
           self
+        rescue ArgumentError
+          self
         end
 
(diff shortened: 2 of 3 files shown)


ShiromMakkad:
Thank you!

PR #1001, opened 2026-06-16
Use OpenSSL for {Ed,X}25519 and ChaCha20-Poly1305

### LINKED ISSUE #983: Missing ivlen for aes128-gcm, aes256-gcm, and chacha20-poly1305 ciphers
When I encrypt the private key with the aes128-gcm@openssh.com  cipher it raises an error when I try to use it with Net:SSH.

```ruby
nil can't be coerced into Integer (TypeError)
              key = BCryptPbkdf::key(password, salt, keylen + ivlen, rounds)
                                                              ^^^^^
```

This call to get the lengths expects three values to return: https://github.com/net-ssh/net-ssh/blob/f430bb5faac6cbae20ffb2b67f87ce9cee74406c/lib/net/ssh/authentication/ed25519.rb#L67

But `.get_lengths` is finding the `klass` in `SSH_TO_CLASS` and is then only returning 2 values here: https://github.com/net-ssh/net-ssh/blob/f430bb5faac6cbae20ffb2b67f87ce9cee74406c/lib/net/ssh/transport/cipher_factory.rb#L103

Reproduction notes: 
OpenSSL version: `OpenSSL 3.6.0`
Made the keypair using: `ssh-keygen -Z aes256-gcm@openssh.com`

bdewater-thatch:
Ruby's openssl gem 3.2.0 (released September 21, 2023) exposes the OpenSSL APIs needed to replace the Ed25519, X25519, and ChaCha20-Poly1305 optional gem implementations when backed by OpenSSL 1.1.1 or another crypto backend providing those primitives.

This requires raising the minimum Ruby version to 2.7 to match the openssl gem requirement. I think this is fine since Ruby 2.7 was released December 25, 2019 and reached EOL on March 30, 2023.

While working on this I ran into earlier reported AEAD-encrypted OpenSSH private key loading, so I also fixed https://github.com/net-ssh/net-ssh/issues/983.

Unit and targeted integration tests pass.


mfazekas:
@bdewater-thatch thanks much, amazing, simplifies so much stuff! 

One minor concern is that MacOS still ships with ruby 2.6, but there is already a warning, so should not be a real concern:
```
WARNING: This version of ruby is included in macOS for compatibility with legacy software. 
In future versions of macOS the ruby runtime will not be available by 
default, and may require you to install an additional package.
```

bdewater:
A quick Google search tells me the warning has been in place since 2019, and the last Ruby update in macOS happened in 2022. I think it's fine to leave the past behind us :)

PR #1003, opened 2026-06-18
Port forwarding buffers entire stream in memory (no backpressure) on slow consumer

gucki:
When forwarding a port, transferring a large amount of data through the tunnel (e.g. piping a multi-GB DB dump) causes Ruby memory to grow roughly to the size of the transfer.

Cause: the forwarder relays bytes between the local socket and the SSH channel with no coupling between read rate and write rate, so whichever leg is "fast in / slow out" buffers unboundedly in the Ruby heap:

channel → socket (reading from remote): update_local_window_size keeps sending CHANNEL_WINDOW_ADJUST regardless of whether the local socket is draining, so the peer keeps flooding while socket output grows.

socket → channel (writing to remote): on_process keeps reading the local socket into the channel's output buffer even when the remote window is exhausted.

Expected: memory stays bounded; the slow side applies backpressure to the fast side.

Environment: net-ssh 7.3.2, Ruby 3.x.

### DIFF
--- a/lib/net/ssh/buffered_io.rb
+++ b/lib/net/ssh/buffered_io.rb
@@ -129,6 +129,12 @@ module Net
         input.to_s
       end
 
+      # Returns the number of bytes currently queued in the output buffer
+      # but not yet written to the underlying IO. Used for backpressure.
+      def pending_write_bytes
+        output.length
+      end
+
       private
 
       #--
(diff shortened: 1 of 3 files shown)


gucki:
The fix was tested in my project and works as expected so far.

PR #1004, opened 2026-06-19
Allow registering public key files without private keys to restrict agent-provided identities

doudou:
This behaviour is ssh's own. If a public key file is provided while there is an agent configured, the public key is solely used to prioritize which public key from the agent is used first. It is critical for agents that propose a lot of keys to avoid key limits imposed by servers (e.g. in Ubuntu)


mfazekas:
This feature landed in main via #942 and #1006 (as `file.end_with?(".pub")` → `:pubkey_file_only`). Feel free to close this one.

doudou:
Understood. Thanks !

